Issue #51, opened 2025-05-03
JOSS review - remarks about code


gdalle:
Some remarks on the code as part of the [JOSS review](https://github.com/openjournals/joss-reviews/issues/8077) for your package:

- The code example in the README fails with a `SingularException` when I run it on SSD v0.1.1 and the latest Julia 1.11.

<details>

```julia
julia> fit!(plds, observations)
Fitting Poisson LDS via LaPlaceEM...   0%|▏                                                 |  ETA: 0:40:45 ( 2.45  s/it)ERROR: TaskFailedException

    nested task error: PosDefException: matrix is not positive definite; Factorization failed.
    Stacktrace:
      [1] cholesky!(F::SparseArrays.CHOLMOD.Factor{…}, A::SparseArrays.CHOLMOD.Sparse{…}; shift::Float64, check::Bool)
        @ SparseArrays.CHOLMOD ~/.julia/juliaup/julia-1.11.5+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/SparseArrays/src/solvers/cholmod.jl:1462
      [2] cholesky!
        @ ~/.julia/juliaup/julia-1.11.5+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/SparseArrays/src/solvers/cholmod.jl:1455 [inlined]
      [3] #cholesky#14
        @ ~/.julia/juliaup/julia-1.11.5+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/SparseArrays/src/solvers/cholmod.jl:1497 [inlined]
      [4] cholesky
        @ ~/.julia/juliaup/julia-1.11.5+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/SparseArrays/src/solvers/cholmod.jl:1490 [inlined]
      [5] cholesky
        @ ~/.julia/juliaup/julia-1.11.5+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/SparseArrays/src/solvers/cholmod.jl:1609 [inlined]
      [6] gaussian_entropy(H::Symmetric{Float64, SparseArrays.SparseMatrixCSC{Float64, Int64}})
        @ StateSpaceDynamics ~/.julia/packages/StateSpaceDynamics/zrsML/src/Utilities.jl:513
      [7] smooth(lds::StateSpaceDynamics.LinearDynamicalSystem{…}, y::Matrix{…}, w::Vector{…})
[...]

PR #53, opened 2025-06-03
Package wide type declaration refactoring 

jfour1e:
Multiple dispatch for the following functions across LDS, SLDS, Mixture Models, HMMs, Preprocessing, etc 
- Sample 
- Loglikelihood
- constructor 
- fit!

### DIFF
--- a/src/EmissionModels.jl
+++ b/src/EmissionModels.jl
@@ -20,10 +20,10 @@ Gaussian Emission Models
 
 GaussianEmission model with mean and covariance.
 """
-mutable struct GaussianEmission <: EmissionModel
+mutable struct GaussianEmission{T<:Float64} <: EmissionModel
     output_dim::Int # dimension of the data
-    μ::Vector{Float64}  # mean 
-    Σ::AbstractMatrix{Float64}  # covariance matrix
+    μ::AbstractVector{T}  # mean 
+    Σ::AbstractMatrix{T}  # covariance matrix
 end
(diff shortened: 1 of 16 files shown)


gdalle:
Hi there!

I was just passing by to see if my JOSS comments were being taken into account, but I'm not sure I understand this PR. Why do you go to such lengths to reduce genericity?
To clarify, when I mentioned in #51 that generic types are a performance pitfall, there is more to be aware of:

- Abstract type annotations for arguments do not change the performance of _functions_: `f(x::AbstractVector{<:Real})` and `f(x::Vector{Float64})` will have the exact same speed
- Abstract type annotations for fields do change the performance of _structs_, but there are ways to keep both generality and speed:

```julia
struct A
    x::AbstractVector{<:Real}  # too slow
end

struct B
    x::Vector{Float64}  # too specific
end

struct C{T<:Real}
    x::Vector{T}  # fast, a little more generic
end

struct D{T<:Real, V<:AbstractVector{T}}
    x::V  # fast, fully generic
end
```

In addition, this PR contains a lot of `T<:Float64`, which doesn't help because `Float64` is a concrete type, so it doesn't have any subtypes except itself. That's why I thought I might contribute a quick feedback, do check out the performance tips section of the manual for more details!

gdalle:
Yes, that would tick all the boxes.

rsenne:
Great!

PR #55, opened 2025-06-09
Switch sample to Random.rand

rsenne:
1. Changes all instances of `sample` to `Random.rand`. 
2.  This now allows all rand functions to accept an RNG that defaults to `default_rng()` if none is provided.
3. Fixes test to use this new API
4. Does not actually fix the singular error this branch originally was meant to fix. Deeper fixes will be necessary--merging back into joss_ and plan on re-opening branch to address


rsenne:
Added a unit test for PPCA sampling. Also, kicks the documentation can down the road, but that's fine as will be resolved in `docs_dev_` branch.

rsenne:
Merging as failing tests are brittleness in EM which is due to poor parameter initialization.

PR #71, opened 2025-08-05
Blue formatting

tomdstone:
I was curious and looked into blue style formatting for Julia as it is one of the badges you have on your README, and work is slow today so I figured I'd go ahead and implement blue formatting. Again I figure its not one of your priorities

### DIFF
--- a/src/EmissionModels.jl
+++ b/src/EmissionModels.jl
@@ -8,7 +8,8 @@ This module implements various emission models for state space modeling, includi
 
 # Exports
 export EmissionModel, RegressionEmission
-export GaussianEmission, GaussianRegressionEmission, BernoulliRegressionEmission, PoissonRegressionEmission, AutoRegressionEmission
+export GaussianEmission, GaussianRegressionEmission, BernoulliRegressionEmission,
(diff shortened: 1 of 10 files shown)


rsenne:
Yes! We do use the blue style. I haven't run the julia formatter in a while. I need to add a check that pull requests adhere to the blue style. Wuuld you be able to add this too?

tomdstone:
Well it looks like `JuliaFormatter.jl` disagrees with the `julia-format` action, so I'm not sure how useful of an action it would be. Would it be too much to just ask folks to run the formatter on their code whenever they open a PR?

rsenne:
Hi Tom! Sorry for not being more clear, that was the idea to make sure users formatted before pushing. Check out this test on [HiddenMarkovModels.jl](https://github.com/gdalle/HiddenMarkovModels.jl/blob/main/test/runtests.jl), line 35-37!

tomdstone:
Ok, that change has been made!

tomdstone:
Notably it does not format the docstrings, I'm of the opinion that that's ok, but its easy enough to change. I had just reformatted them manually myself.

tomdstone:
Nope! Good to go

PR #91, opened 2026-05-26
BTD path go brrrr

rsenne:
Basically just lots of overhauls to everything. Copilot didn't even get it all lol. Some things to note:

- Change in params (mentioned via email)
- Currently control matrices in the state and obs paths are gated by the corresponding matrices (i.e., A and C) might want to separate out the m-step updates if someone wants to use a fixed observation matrix or dynamics matrix?
- Need to determien what we want to do with Kalman path
- Probably other things

This pull request adds a suite of new benchmarking and diagnostic scripts to the repository, focused on evaluating and profiling the performance and correctness of the state-space modeling and inference routines (notably for Gaussian and Poisson linear dynamical systems). It also updates the documentation and example usage to match recent API changes, particularly the renaming of parameters in the observation models. The new scripts provide tools for allocation profiling, benchmarking block tridiagonal solvers, comparing hand-rolled and Optim-based Newton solvers, investigating ELBO vs. marginal log-likelihood, and large-scale performance comparisons.

Key changes include:

**New benchmarking and diagnostic scripts:**

* Added `benchmarking/alloc_profile.jl` for detailed allocation profiling of the block tridiagonal (BTD) path, to identify remaining memory allocations after the PDMats refactor.
* Added `benchmarking/bt_solve_bench.jl` to benchmark the block tridiagonal solver and compare it to sparse direct solvers (UMFPACK), focusing on Hessian construction and solve times for Poisson LDS.
* Added `benchmarking/bt_vs_optim_sweep.jl` to compare the performance (time, memory) of the hand-rolled Newton solver versus Optim.jl's Newton solver with sparse Hessians, across a range of problem sizes.
* Added `benchmarking/optim_vs_handrolled.jl` for a direct performance comparison between the package's custom Newton solver and Optim.jl's Newton solver on a small Poisson LDS problem.
* Added `benchmarking/elbo_investigation.jl` to investigate the difference between the ELBO and the true marginal log-likelihood during EM on a Gaussian LDS, providing insight into convergence and constant offsets.
* Added `benchmarking/large_td_vs_kalman.jl` to benchmark large-scale fits comparing the block tridiagonal (TD) approach and the Kalman filter-based approach on high-dimensional problems.

**Documentation and example updates:**

* Updated `README.md` to match recent API changes: renamed `log_d` to `d` in `PoissonObservationModel`, added missing bias parameters, updated the data generation and fitting examples, and clarified the meaning of the `d` parameter.

**Test and workflow adjustments:**

* Updated `.github/workflows/benchmark.yml` to use the new `d` parameter name (instead of `log_d`) in `PoissonObservationModel` to match the refactored API.

These changes collectively improve the ability to benchmark, profile, and validate the performance and correctness of the state-space inference algorithms, especially after recent structural and API changes.


rsenne:
Note to self: This (and future PRs) should be squash merged for linear history now that updates past this point should be a bit more surgical than this

rsenne:
Yikes need to add more tests for sure (my bad)

src/StateSpaceDynamics.jl
```diff
@@ -2,28 +2,25 @@ module StateSpaceDynamics
...
-using LineSearches
-using Optim
 using PDMats
-using ProgressMeter
 using Random
 using SparseArrays
-using SpecialFunctions
-using Statistics
+
+using Optim: Optim, optimize, LBFGS, HagerZhang
```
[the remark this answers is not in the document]
rsenne: this is wrong. resolving

rsenne:
@harrisonritz had a chance to review yet? (no rush just checking in)

harrisonritz:
Not yet, at a conference.

Are you feeling confident, or do you want me to take a pass through it later this week?

rsenne:
Oh right sorry! No worries, like i said no rush just wanted to check in

src/models/lds/gaussian.jl
```diff
@@ -14,39 +15,75 @@ end
...
+    # (it consumes `x_smooth` / `p_smooth` / `p_smooth_tt1` directly), and
+    # the smoother aliases `p_smooth` / `p_smooth_tt1` to
+    # `sws.p_smooth_shared` on every E-step — so all four `(D, D, tsteps)`
+    # arrays are allocated then either ignored or immediately overwritten.
+    # At `(D=128, T=250, N=500)` that's ≈ 64 GB of pure waste.
+    #
+    # Default (`cov_alias=false`) preserves the original layout: SLDS /
+    # Poisson / ragged / single-trial paths write into the per-trial
+    # `p_smooth` and may invoke the legacy `sufficient_statistics!(fs)`
+    # which populates `E_zz` / `E_zz_prev`.
```
harrisonritz: I think that `cov_alias` needs a more informative name. How about `smooth_cov=:epoch` vs `smooth_cov=:global`?

src/models/lds/gaussian.jl
```diff
@@ -14,39 +15,75 @@ end
...
 
 function _extract_obs_params(obs_model::GaussianObservationModel{T}) where {T}
-    return (C=obs_model.C, R=obs_model.R, d=obs_model.d)
+    return (C=obs_model.C, R=obs_model.R, d=obs_model.d, D=obs_model.D)
 end
 
 function _extract_obs_params(obs_model::PoissonObservationModel{T}) where {T}
-    return (C=obs_model.C, log_d=obs_model.log_d, d=exp.(obs_model.log_d))
+    return (C=obs_model.C, d=obs_model.d)
 end
```
harrisonritz: I don't think we should keep obs_model.d or state_model.b.
I know its conceptually simple for mean-subtraction, but (intercept-only) D generalizes this with little additional computational cost.

src/models/lds/gaussian.jl
```diff
@@ -67,39 +104,71 @@ function _get_all_params_vec(
...
+    state_params,
+    obs_params,
+    obs_model::PoissonObservationModel,
+    u_trial::AbstractMatrix,
+    v_trial::AbstractMatrix,
 )
     tsteps = size(x_trial, 2)
+    # Poisson obs model has no D matrix; v_trial is accepted for signature
+    # parity with the Gaussian path but must be empty (validated by callers).
+    @assert size(v_trial, 1) == 0 "Poisson observation model does not support obs inputs"
```
harrisonritz: why not have observation inputs for poisson model?

src/models/lds/gaussian.jl
```diff
@@ -118,32 +192,50 @@ function _sample_trial!(
...
+    v_trial = _check_obs_control(obs_control_seq, lds.obs_input_dim, Ti, lds.obs_model)
 
-    x = Matrix{T}(undef, lds.latent_dim, Int(tsteps))
-    y = Matrix{T}(undef, lds.obs_dim, Int(tsteps))
-    _sample_trial!(rng, x, y, state_params, obs_params, lds.obs_model)
+    x = Matrix{T}(undef, lds.latent_dim, Ti)
+    y = Matrix{T}(undef, lds.obs_dim, Ti)
+    _sample_trial!(rng, x, y, state_params, obs_params, lds.obs_model, u_trial, v_trial)
     return x, y
 end
```
harrisonritz: I think that `control_seq` has the wrong semantics, even though these are sometimes called controls. I think that `inputs` are a better label than `controls`. how about `state_input_seq` vs `obs_input_seq`?

src/models/lds/gaussian.jl
```diff
@@ -157,27 +249,155 @@ function Random.rand(
...
+
+# ============================================================================
+# Control-sequence normalization helpers. The public `control_seq`/`obs_control_seq`
+# kwargs accept either `nothing` (no inputs — must match a zero-column `B`/`D`)
+# or per-trial matrices. Internally every sampler/smoother/M-step expects an
+# `AbstractMatrix{T}` of shape `(u_dim, T_i)` (possibly `0 × T_i`), so these
+# helpers canonicalize on the way in.
+# ============================================================================
+
+function _check_control(cs::Nothing, expected_dim::Int, tsteps::Int, name::AbstractString)
```
harrisonritz: re: above, `_check_state_inputs`

src/models/lds/gaussian.jl
```diff
@@ -157,27 +249,155 @@ function Random.rand(
...
+function Random.rand(lds::LinearDynamicalSystem, tsteps::Integer; kwargs...)
+    return rand(Random.default_rng(), lds, tsteps; kwargs...)
 end
 
 function Random.rand(
-    lds::LinearDynamicalSystem, tsteps_per_trial::AbstractVector{<:Integer}
+    lds::LinearDynamicalSystem, tsteps_per_trial::AbstractVector{<:Integer}; kwargs...
+)
+    return rand(Random.default_rng(), lds, tsteps_per_trial; kwargs...)
+end
```
harrisonritz: consider splitting off into new script called `simulate`, `sample`, or `rand`?

src/models/lds/gaussian.jl
```diff
@@ -157,27 +249,155 @@ function Random.rand(
...
+
+@inline function _normalize_multitrial_obs_control(
+    cs::Nothing, expected_dim::Int, tsteps_per_trial, ::Type{T}, ::PoissonObservationModel
+) where {T<:Real}
+    expected_dim == 0 || error(
+        "Poisson observation model does not support obs_control_seq (expected_dim must be 0)",
+    )
+    return [zeros(T, 0, Int(Ti)) for Ti in tsteps_per_trial]
 end
```
harrisonritz: consider splitting off into new script called `validate`?

src/gaussian.jl
```diff
@@ -191,17 +411,21 @@ function loglikelihood!(
...
     x0 = lds.state_model.x0
     d = lds.obs_model.d
+    D_obs = lds.obs_model.D
 
-    R_U = UpperTriangular(ws.R_chol_U)
-    Q_U = UpperTriangular(ws.Q_chol_U)
-    P0_U = UpperTriangular(ws.P0_chol_U)
+    R_U = ws.R_PD[].chol.U
+    Q_U = ws.Q_PD[].chol.U
+    P0_U = ws.P0_PD[].chol.U
```
harrisonritz, line 35: do we need to materialize all of these?

src/gaussian.jl
```diff
@@ -191,17 +411,21 @@ function loglikelihood!(
     x::AbstractMatrix{T},
     lds::LinearDynamicalSystem{T0,S,O},
     y::AbstractMatrix{T0},
+    u::AbstractMatrix{T0},
+    v::AbstractMatrix{T0},
 ) where {T<:Real,T0<:Real,S<:GaussianStateModel{T0},O<:GaussianObservationModel{T0}}
```
harrisonritz, line 23: where is `loglikelihood!` called?

src/gaussian.jl
```diff
@@ -1288,6 +2562,10 @@ Fit a Gaussian Linear Dynamical System via Expectation-Maximization.
 - `max_iter::Int=100`: maximum EM iterations
 - `tol::Float64=1e-6`: convergence tolerance on ELBO change
 - `progress::Bool=true`: show progress bar
+- `control_seq`: optional dynamics-input sequence. `Vector{<:AbstractMatrix}`
+  for multi-trial (each `(u_dim, T_i)`); required when `size(state_model.B, 2) > 0`.
+- `obs_control_seq`: optional observation-input sequence (same shape) for the
+  obs-side input matrix `D`. Required when `size(obs_model.D, 2) > 0`.
```
harrisonritz, line 1180: see above re: state_input_seq / obs_input_seq

src/models/lds/gaussian.jl
```diff
@@ -1297,11 +2575,17 @@ function fit!(
...
-    return _fit!(lds, y, max_iter, tol, progress, u0, u, d, Val(lds.kalman_filter))
+    tsteps_per_trial = [size(yt, 2) for yt in y]
+    u_seq = _normalize_multitrial_control(
+        control_seq, lds.state_input_dim, tsteps_per_trial, T, "control_seq"
+    )
+    v_seq = _normalize_multitrial_obs_control(
+        obs_control_seq, lds.obs_input_dim, tsteps_per_trial, T, lds.obs_model
+    )
+    return _fit!(lds, y, max_iter, tol, progress, u_seq, v_seq, Val(lds.kalman_filter))
 end
```
harrisonritz: remove kalman path

harrisonritz:
A couple of high-level comments:
- now that we focus on LDS, does it make sense to just have models/lds, which also contains SLDS? Maybe more folders/fewer folders

- maybe Gaussian.jl --> lds.jl?

- there is a filter_logliklihood function in Gaussian.. doesn't need to be optimized if it's just for comparing against null models. There's also Loglikelihood! (Elbo IIRC), which doesn't seem to get called at all (Q_* functions seem to get used).


I added some comments just as checks, feel free to assign. I also pushed a carelessly (had to run for my bus), so I might have added D to a poisson function 🙊

rsenne:
All good--fixed the Poisson stuff! Re your thoughts:

1.  I'm not sure i totally understand. Do you mean like an LDS folder and an SLDS folder? As SLDS is inside of LDS models already?
2. same as above not sure i totally understand
3. hmm okay, my memory is probably failing me then. Nonetheless then maybe we can make likelihood just be the filter loglikelihood

src/models/lds/gaussian.jl
harrisonritz: or even better: `state_inputs`, `obs_inputs`, and `init_inputs`.

harrisonritz:
> All good--fixed the Poisson stuff! Re your thoughts:
> 
> 1. I'm not sure i totally understand. Do you mean like an LDS folder and an SLDS folder? As SLDS is inside of LDS models already?

If the package now only covers lds, then it seems unnecessary to only have a `src/models/lds/` folder. Either have more differentiation ( `src/models/lds`, `src/models/slds`, `src/models/rslds`, etc) or less differentiation (e.g., `src/lds/...`). Maybe you want to future-proof -- this is minor comment, trying to point out potential inertia.

> 2. same as above not sure i totally understand

ah, remembering now that its differentiating between gaussian / poisson, as much as lds vs slds.
Since this script is the whole fitting procedure, and eg not just the gaussian steps, I think its be more informative to have script names that make it easy for a user to look-up our algorithms. How about `gaussianLDS.jl` vs `poissonLDS.jl` vs `switchingLDS.jl` (ignoring the fact that slds can be either, and poisson has gaussian latent dynamics).

> 3. hmm okay, my memory is probably failing me then. Nonetheless then maybe we can make likelihood just be the filter loglikelihood

How about `Loglikelihood!` becomes `joint_loglikliehood`, and `filter_loglikelihood` becomes `marginal_loglikelihood`? can rearrange to put them together at the bottom; against, afaict, neither are called in the EM procedure.

4. I think we should get rid of b/d/x0, and doing everything through B@state_inputs, D@obs_inputs, and B0@init_inputs (or u, v, u0, if we need short-forms). The time-varying predictors generalize the mean-only case, the mean-only case is an easy default, and having all the inputs together is helpful for things like collinearity checks. 
For the initial condition, I think that the residual cov you'd get from the OLS m-step is equivalent to [Ghahramani & Hintons'](https://www.cs.toronto.edu/~hinton/absps/tr-96-2.pdf) equations for the init cov (but don't remember the proof):

Screenshot 2026-06-05 at 10 39 34 AM

(where $P_{t} = V_{t|1:T} + x_{t|1:T}x_{t|1:T}^{\top}$)

rsenne:
> Maybe you want to future-proof

That was the original motivation--but point well taken--its a bit unintuitive ATM. I prefer `models/lds` and `models/SLDS` as you've suggested

> How about gaussianLDS.jl vs poissonLDS.jl vs switchingLDS.jl

Open to something like this. I'm thinking maybe instead we have: `CommonLDS.jl` / `GaussianState.jl` or something that unifies the state aspect being identical? Then `PoissonLDS.jl`/`GaussianLDS.jl` contain things pertinent to observation model things? Feels like there are a few ways we can slice this.

> `Loglikelihood!` becomes `joint_loglikliehood`, and `filter_loglikelihood` becomes `marginal_loglikelihood`

`loglikelihood` to `joint_loglikelihood!` seems well justified. I could see filter `filter_loglikelihood` becoming `marginal_loglikelihood` or `loglikelihood` as i imagine when people think the "loglikelihood of the LDS" they mean the marginal. The only issue is presents is now GaussianLDS has a special method no other one has e.g., as we can't get the ML of the PoissonLDS we have three options I can see:

1. We make a special method for Gaussians but throw an error if someone calls this method for any other model with an error saying to compute the ELBO or something.
2. We could provide a monte carlo estimate of the LL via a particle filter and then document it being approx
3. We make everything called the elbo and document that this is equivalent to the ML in the case of a gaussian LDS. (also here is a MWE of this point I just made showing in theory they are the same https://pluto.land/n/lj2cxgv4)

> I think we should get rid of b/d/x0

What do you propose as the new parameter names? My one thought is b/d are fairly common parameter names in the literature for these specific biases. x0/p0 are also reasonably common. That's not to say I'm not on board--i see the argument you're making--mostly am curious how we can make this as clean as possible. I guess the things I'm curious to know:

1. Is there a way to keep the common parameter names but just have the the learning be through the control interface as you propose? I guess the init_inputs are always going to be a column of ones?
2. How does this intersect with the control-inputs-present vs. -absent cases? If b becomes "the constant column of u," then the no-control case still needs that column of ones
3. Can you remind me how x0 is handled?
4. Do you want to do this here or in the next PR?

harrisonritz:
> > Maybe you want to future-proof
> 
> That was the original motivation--but point well taken--its a bit unintuitive ATM. I prefer `models/lds` and `models/SLDS` as you've suggested

Up to you, minor concern.

> > How about gaussianLDS.jl vs poissonLDS.jl vs switchingLDS.jl
> 
> Open to something like this. I'm thinking maybe instead we have: `CommonLDS.jl` / `GaussianState.jl` or something that unifies the state aspect being identical? Then `PoissonLDS.jl`/`GaussianLDS.jl` contain things pertinent to observation model things? Feels like there are a few ways we can slice this.

Am I right in thinking that `GaussianState.jl` would have the BTD algorithm, and `PoissonLDS.jl`/`GaussianLDS.jl` would focus on the msteps? In that case, `estep.jl` / `mstep.jl` scripts might be more straightforward.

> > `Loglikelihood!` becomes `joint_loglikliehood`, and `filter_loglikelihood` becomes `marginal_loglikelihood`
> 
> `loglikelihood` to `joint_loglikelihood!` seems well justified. I could see filter `filter_loglikelihood` becoming `marginal_loglikelihood` or `loglikelihood` as i imagine when people think the "loglikelihood of the LDS" they mean the marginal. The only issue is presents is now GaussianLDS has a special method no other one has e.g., as we can't get the ML of the PoissonLDS we have three options I can see:

IMO better to have `marginal_loglikelihood` that just `loglikelihood`. Keeps it precise, and the clarification can be in the doc string.

> 1. We make a special method for Gaussians but throw an error if someone calls this method for any other model with an error saying to compute the ELBO or something.
> 2. We could provide a monte carlo estimate of the LL via a particle filter and then document it being approx
> 3. We make everything called the elbo and document that this is equivalent to the ML in the case of a gaussian LDS. (also here is a MWE of this point I just made showing in theory they are the same https://pluto.land/n/lj2cxgv4)

I think (1) makes the most sense for now, and then put (2) on the roadmap. I think that it's worth have a proper marginal loglik for doing model comparison. (3) is really nice, but you said there was an offset -- is this when it's higher dimensional?


> 
> > I think we should get rid of b/d/x0
> 
> What do you propose as the new parameter names? My one thought is b/d are fairly common parameter names in the literature for these specific biases. x0/p0 are also reasonably common. That's not to say I'm not on board--i see the argument you're making--mostly am curious how we can make this as clean as possible. I guess the things I'm curious to know:
> 
> 1. Is there a way to keep the common parameter names but just have the the learning be through the control interface as you propose? I guess the init_inputs are always going to be a column of ones?

If I get what you're saying, one option would be to never use b/d/x0 during fitting, but to save b,d,x0 fields using B[:,1], D[:,1], B0[:,1]. This just risks confusing people, and we could merely point to B[:,1], D[:,1], B0[:,1] instead.

`init_inputs` can be any predictors x epochs matrix, it's just an input like anything else. For example, if modelling decision-making, you might include the previous choice in the initial conditions (in our paper, we included previous task and previous RT).

> 2. How does this intersect with the control-inputs-present vs. -absent cases? If b becomes "the constant column of u," then the no-control case still needs that column of ones

If there's no control/inputs, then I don't think there should be a `b` either. We can carve out a case where theres no inputs at all, 

> 3. Can you remind me how x0 is handled?

The initial condition are set by the inputs
```Julia
# pred_mean: states x steps x epochs
# B0: states x predictors
# init_inputs: predictors x epochs
pred_mean[:,1,:] = B0 * init_inputs
```

And then the update is done using OLS just like B:
```julia
B0 = smooth_mean[:,1,:] / init_inputs 
```
 With P0 updated the same way as Q/R.
 
 One important thing this adds is that it lets you avoid constant inputs. e.g., if you want to have separate inputs for different sub-sections, or if you want to use a spline basis to capture a time-varying inputs. By intentionally including the intercept, folks will have to think about the collinearity of their predictors.

> 4. Do you want to do this here or in the next PR?

I'm ok pushing to next PR, but I think that a (b,d,x0) --> (B,D,B0) should be in this major release. 
It is breaking enough that the main and dev will really start to drift.

rsenne:
> Up to you, minor concern.

Agreed it's minor. I'll sit on it. Pretty inconsequential while LDS-style models are the only thing in the namespace; easy to revisit if something non-LDS lands.

> Am I right in thinking that GaussianState.jl would have the BTD algorithm, and PoissonLDS.jl/GaussianLDS.jl would focus on the msteps?

same as above--I can see how in the future this would need to change again. E.g., if SLDS is inside LDS then where should the E/M step go? I do agree soem re-org is necessary.

> I think (1) makes the most sense for now, and then put (2) on the roadmap. I think that it's worth have a proper marginal loglik for doing model comparison. (3) is really nice, but you said there was an offset -- is this when it's higher dimensional?

Sounds good to me. RE: 2 I have some particle filtering stuff I'm doing elsewhere i can put this in. 

The one I meant is just an additive constant, same sense that least-squares and the Gaussian LL differ by the normalizer terms; they track each other exactly up to that constant, no dimension dependence. The ELBO and the exact marginal LL coincide for a Gaussian LDS because q is the exact posterior (KL = 0); that's the ln p(X) = L(q,θ) + KL(q‖p) decomposition, see Bishop 9.4, e.g., 9.70–9.74 (~p.450–452)

> If I get what you're saying, one option would be to never use b/d/x0 during fitting, but to save b,d,x0 fields using B[:,1], D[:,1], B0[:,1]. This just risks confusing people, and we could merely point to B[:,1], D[:,1], B0[:,1] instead.

My proposal: the affine params stay first-class in the public API, while all the fitting machinery moves private. Internally b can just be B[:,1].  I'm not arguing against unifying the M-step, that part I'm fully on board with. But I'd keep b/d/x0 as named, exposed fields rather than telling users "the offset is column 1 of B." I guess analogously in linear regression the intercept is mechanically just the coefficient on a constant column of the design matrix but you'd generally export that as a separate field, even if just for convenience.

I think my reason for hesitancy on these points mostly comes down to different use cases--ideally one can do something as complicated as your recent paper with SSD and as easy as fitting an LDS with affine terms if someone wanted to. So as long as we can have our cake and eat it too here I'd be happy. 

Same somewhat goes for the x0/P0. I like the idea of generalizing the initial distribution but in the case where there is no input e.g., i just need to estimate the initial distribution and there is no input--how is this h> Up to you, minor concern.

Agreed it's minor. I'll sit on it. Pretty inconsequential while LDS-style models are the only thing in the namespace; easy to revisit if something non-LDS lands.

> Am I right in thinking that GaussianState.jl would have the BTD algorithm, and PoissonLDS.jl/GaussianLDS.jl would focus on the msteps?

same as above--I can see how in the future this would need to change again. E.g., if SLDS is inside LDS then where should the E/M step go? I do agree some re-org is necessary. Maybe i am over complicating this

> I think (1) makes the most sense for now, and then put (2) on the roadmap. I think that it's worth have a proper marginal loglik for doing model comparison. (3) is really nice, but you said there was an offset -- is this when it's higher dimensional?

Sounds good to me. RE: 2 I have some particle filtering stuff I'm doing elsewhere i can put this in. 

The one I meant is just an additive constant, same sense that least-squares and the Gaussian LL differ by the normalizer terms; they track each other exactly up to that constant, no dimension dependence. The ELBO and the exact marginal LL coincide for a Gaussian LDS because q is the exact posterior (KL = 0); that's the ln p(X) = L(q,θ) + KL(q‖p) decomposition, see Bishop 9.4, e.g., 9.70–9.74 (~p.450–452)

> If I get what you're saying, one option would be to never use b/d/x0 during fitting, but to save b,d,x0 fields using B[:,1], D[:,1], B0[:,1]. This just risks confusing people, and we could merely point to B[:,1], D[:,1], B0[:,1] instead.

My proposal: the affine params stay first-class in the public API, while all the fitting machinery moves private. Internally b can just be B[:,1].  I'm not arguing against unifying the M-step, that part I'm fully on board with. But I'd keep b/d/x0 as named, exposed fields rather than telling users "the offset is column 1 of B." I guess analogously in linear regression the intercept is mechanically just the coefficient on a constant column of the design matrix but you'd generally export that as a separate field, even if just for convenience

I think my reason for hesitancy on these points mostly comes down to different use cases--ideally one can do something as complicated as your recent paper with SSD and as easy as fitting an LDS with affine terms if someone wanted to. So as long as we can have our cake and eat it too here I'd be happy. 

Same somewhat goes for the x0/P0. I like the idea of generalizing the initial distribution but in the case where there is no input e.g., i just need to estimate the initial distribution and there is no input--what does `B0 * init_inputs` reduce to?

My suggestion is we cut a release based on whats left that's low-hanging and then we can continue to do smaller targeted PRs (e.g., we can finalize an API for control, PF, etc.) handled?

My suggestion is we cut a release based on whats left that's low-hanging and then we can continue to do smaller targeted PRs (e.g., we can finalize an API for control, PF, etc.) My hope is to start cutting releases more frequently so I'd imagine this wouldn't be in the lurch for to long

Also if this ends up being easier to discuss over a zoom call lmk lol

harrisonritz:
I think there was a copy error, so I'll break the quote-tree.
For the most part, this sounds good.

- agree constant offset doesn't matter for comparing within a model, but it will matter for comparing between models (eg a VAR(1) that uses a MVN at the sensor level). I think we're on the same page on this. Great that you verified the equivalence, thats super cool.

- I don't think that a constant intercept should not be required (eg it will be collinear with splines). Maybe a good compromise would be to have an `input_format` argument that structured like:
```julia
 input_format = (state="matrix", observation="constant", init="constant")
 ```
Defaults to all "constant", and "matrix" doesn't pre-populate the intercept. Internally, we convert this to B/D/B0 matrices. If user asked for "constant", we return b/B with B all zeros, if they ask for "matrix", we also return b/B, with b all zeros.

IDK about release, your call. If there's not a huge user base, then a few semi-breaking release back-to-back won't be a huge deal.

**Comment by rsenne** [comment]
Sorry--I think i didn't make something clear from before--part of my thinking was to loosen how `fit_bool` (I'm also in favor of redoing fit_bool in some way that makes it easier to see what parameters to fit but thats a discussion for a future PR), and consequently the M-step would work.

Right now--the way I handle intercepts is to add a column of ones to the A matrix and then do the OLS update,--but this isn't needed and we can instead add it to the B matrix like you said with an input vector of ones. This might make my proposal clearer--e.g., bias can be optional, just like inputs are and these details can be hashed out in fit-bool. Does this make sense?

As for user base--its probably less than 15-20 people. The biggest reason for release was to make it clear we are still devving/getting into the practice of making more frequent releases as necessary but it can go either way. I think at the very least we should merge into main soon so that we can make more self-contained changes

**Comment by harrisonritz** [comment]
Makes sense, and sounds like a good plan :)

**Review by harrisonritz — APPROVED** [review]

**Comment by rsenne** [comment]
For some reason I didn't see any of these comments above until now? Do you want me to go through them? Or should we start fresh on a PR into main?

**Comment by harrisonritz** [comment]
Ohhh, that must be the review mechanism, sorry. Yes, take a look, some are pretty minor. Feel free to assign

File: src/models/lds/gaussian.jl (hunk 4 above)
Comment (rsenne) [reply src/models/lds/gaussian.jl]: No reason other than I hadn't gotten to it/didn't know if needed for this current refactor. Was going to open an issue on it

**Comment by rsenne** [comment]
Hey @harrisonritz i started a major re-org. Let me know what you think

File: src/gaussian.jl (hunk 11 above)
Comment (harrisonritz) [reply src/gaussian.jl:1180]: even better: state_inputs and obs_inputs

**Comment by harrisonritz** [comment]
Updates look good!
A few more thoughts:

---
(1) I played around with renaming the marginal loglik functions `marginal_loglikelihood` internally, and providing the option for an alias, e.g., 
```julia
const loglikelihood = marginal_loglikelihood
```
But this doesn't play well with the StatsAPI import. Is this import for functionality, or just to provide a similar method to other packages?

e.g., another aliasing option could be something like:
```julia
function loglikelihood(
     lds::LinearDynamicalSystem{T,S,O}, y::AbstractArray{Union{T, AbstractArray{T}}}
     ) where {T<:Real,S<:AbstractStateModel{T},O<:AbstractObservationModel{T}}
    return marginal_loglikelihood(lds, y)
end
```

I think for internals, it's awkward to have joint_loglik and loglik, and an alias would still provide user-facing simplicity & statsAPI interoperability.

---
(2) A little awkward that (the renamed) latent.jl is common to all algorithms, but isn't included in common.jl. Was surprised by the contents in both scripts. 

Maybe worth merging `core` into `models/lds`, and adding the functions in `common.jl`  to `Utilities.jl`? side note: should have consistency in capitalization (though don't love Julia's style guide -- find snake case most readable).

Getting into the weeds, sorry, but should be easy fix.

rsenne:
> Is this import for functionality, or just to provide a similar method to other packages?

StatsAPI is a pretty large effort to unify what methods are called what in packages. Its something I've been meaning to do as it does make package interoperability easier in many cases and it has a nice benefit of standardizing the ecosystem. Other methods that should has the StatsAPI flag is also `fit!`. 

I renamed the methods to `loglikelihood` as I had this realization while working on it, and many other packages use `loglikelihood`, even for LVMs. I do agree `marginal_loglikelihood` is more specific. Not certain how i feel about using an alias. If the idea is to differentiate the backend (i.e., to make coding easier) maybe `marginal_loglikelihood` should be the alias of `loglikelihood` as opposed to the opposite? 

>  A little awkward that (the renamed) latent.jl is common to all algorithms, but isn't included in common.jl. Was surprised by the contents in both scripts. 

Finding the right homes for everything has been somewhat challenging. But these are all good points. Let me continue to make some changes

rsenne:
IF these tests pass I am going to squash merge into dev_ as this is now the de_facto branch

Issue #105, opened 2026-07-02
Fix JET


rsenne:
Something in the Kalman path is making it unhappy

harrisonritz:
Addressed in https://github.com/depasquale-lab/StateSpaceDynamics.jl/pull/124

rsenne:
Closing as completed in #124

PR #124, opened 2026-07-04
Resolve JET issues

harrisonritz:
**summary**: Add explicit `::SubArray{T}` assertions to `view` slices in the LDS, Kalman, and sufficient-statistics paths so JET no longer infers spurious `Any`-typed branches for BLAS and related calls. This re-enables `JET.test_package` on Julia 1.11-1.12 while still skipping 1.13+ because of a JET-internal collector bug, and also drops a redundant matrix `estep!` overload.

One key issue was that `estep!` with matrix-typed `y` is never called (ie no `_fit_tridiag!` for Matrix-typed `y`). For some reason this causes things like JuliaFormatter to crash.

Matrix-typed `estep!` is removed for now. There's still matrix-typed methods at lower levels, e.g., `smooth!`. They might be faster for matched-length epochs, eh? So can decide whether to add a matrix-typed `_fit_tridiag!` method.

Big enough change that I didn't want to hotfix to main. Note that I'm testing at `1.13.0-rc1`.

Addresses issue https://github.com/depasquale-lab/StateSpaceDynamics.jl/issues/105


harrisonritz:
@rsenne this one is simple enough to look at right away.

test/runtests.jl
```diff
@@ -41,37 +41,25 @@ using SSDTest
...
+            if v"1.11" <= VERSION < v"1.13.0-"
+                # Static analysis over the package's own modules. The union-split
+                # false positives that used to block this — a workspace
+                # `@views`/`view` over a field typed `Array{T,3}` / `Matrix{T}` /
+                # `Vector{PDMat{T,Matrix{T}}}` inferring a spurious
+                # `Union{SubArray{Any,…}, SubArray{T,…}}` whose `Any` branch
+                # failed to match `BLAS.ger!` / `BLAS.syrk!` / `X_A_Xt` /
+                # `logpdf` / `Symmetrize!` — are cleared by asserting each view
+                # result `::SubArray{T}` at the call site (see the NOTE atop
+                # `kalman.jl` / `sufficient_statistics.jl`).
```
rsenne:
```suggestion
```

test/runtests.jl
```diff
@@ -41,37 +41,25 @@ using SSDTest
...
-                    StateSpaceDynamics; target_modules=(StateSpaceDynamics,)
-                )
+            # Skipped on Julia 1.13+ (the `-` in `v"1.13.0-"` also excludes
+            # pre-releases such as `1.13.0-rc1`): JET 0.11.5's report collector
+            # leaves `#undef` slots in its internal `Vector{InferenceErrorReport}`
+            # once the analysis produces many reports (~2200 raw, mostly
+            # Base/stdlib), so `unique!` throws `UndefRefError` *before*
+            # `target_modules` filtering. That is a JET-internal bug on 1.13,
+            # unrelated to this package — our own analysis is clean (0 reports).
+            # Raise the upper bound once a JET release supports 1.13.
```
rsenne:
```suggestion
# skip pre-release
```

rsenne: In general, I'd like to avoid these types of very verbose comments--I feel they add too much bloat. In general, explanation is good but I think we can probably minimize this much detail. Also for block quotes: #= my comment =# is more idiomatic

test/runtests.jl
```diff
@@ -41,37 +41,25 @@ using SSDTest
...
+            if v"1.11" <= VERSION < v"1.13.0-"
```
rsenne: We should be able to do something like this:

```
if v"1.11" <= VERSION && isempty(VERSION.prerelease)
```

modulo syntax correctness (didn't check lol)

src/stats/sufficient_statistics.jl
```diff
@@ -1,3 +1,11 @@
+# NOTE (JET view-eltype union): under JET's generic package analysis a workspace
+# field typed `Matrix{T}` / `Array{T,3}` (with `T` a free `<:Real` typevar) makes
+# `view(field, …)` infer as `Union{SubArray{Any,…}, SubArray{T,…}}`. The spurious
+# `Any` branch fails to match the strict `BLAS.ger!` / `BLAS.syrk!` / `X_A_Xt`
+# signatures, producing false-positive `no matching method` reports. Asserting the
+# `view` result `::SubArray{T}` intersects away the `Any` branch (the concrete
+# `ndims` from the real branch is preserved), matching runtime types exactly.
```
rsenne:
```suggestion
```

src/lds/fit_LDS.jl
```diff
@@ -829,7 +832,9 @@ function _smooth_mean_only_batched!(
     x_flat = reshape(sws.batched_x_mat, n_active, ntrials)
     @. grad_flat = -grad_flat
 
-    neg_sub_v = view(sws.btd.neg_sub, 1:(tsteps - 1))
+    # `::SubArray{Matrix{T}}` narrows JET's `view`-eltype union over the
+    # `Vector{Matrix{T}}` field so `block_tridiagonal_backsubst!` matches.
```
rsenne:
```suggestion
```

src/lds/fit_LDS.jl
```diff
@@ -600,8 +600,11 @@ function _precompute_shared_cov!(
         p_smooth_v, p_smooth_tt1_v, neg_sub_v, neg_diag_v, neg_super_v, btd
     )
 
-    @views for i in 1:tsteps
-        Symmetrize!(p_smooth_shared[:, :, i])
+    # `::SubArray{T}` narrows JET's `view`-eltype union over the `Array{T,3}`
+    # workspace field so `Symmetrize!`'s typed signature matches (cf. the NOTE
+    # in kalman.jl / sufficient_statistics.jl).
```
rsenne:
```suggestion
```

src/stats/kalman.jl
```diff
@@ -6,6 +6,14 @@
 # backend — all Gaussian fitting goes through the block-tridiagonal path.
 # Ported from StateSpaceAnalysis.
 #
+# NOTE (JET view-eltype union): under JET's generic package analysis a workspace
+# field typed `Array{T,3}` / `Matrix{T}` (with `T` a free `<:Real` typevar) makes
+# `@views`/`view` over it infer as `Union{SubArray{Any,…}, SubArray{T,…}}`. The
+# spurious `Any` branch fails to match strict `X_A_Xt` / `BLAS.syrk!` / `logpdf`
+# signatures, yielding false-positive `no matching method` reports. Asserting the
+# view result `::SubArray{T}` intersects away the `Any` branch (the concrete
+# `ndims` from the real branch is preserved), matching runtime types exactly.
```
rsenne:
```suggestion
```

rsenne (changes requested):

rsenne:
After reviewing this I'd like to suggest a quick re-work if that's amenable. I think the amount of rewritten code means we should have a small inline utility we can grep when this is fixed upstream if possible. I'm thinking this can be a repalcement

```julia
#=
Workaround for JET union-split false positive on views with unbound eltype
(remove when fixed upstream, see: JET issue here)
=#
@inline tview(A::AbstractArray{T}, I...) where {T} = view(A, I...)::SubArray{T}
```

Then we can put this in one place, remove the related comments, keeps it clean etc.

I boiled this down to the following MWE that can be submitted upstream to JET as an issue as I perused their issues and couldn't find one so I think this is genuinely unknown (though I could be wrong). I can submit or you can your call!

```julia
# MWE: Version Info Julia 1.12.6, JET 0.11.5
using JET

# any eltype-strict callee triggers it (BLAS.syrk!, Symmetrize!, logpdf, ...)
strict!(S::AbstractMatrix{T}) where {T<:Real} = (S .+= one(T); nothing)

f(G::Array{T,3}) where {T<:Real} = strict!(view(G, :, :, 1))

#errors
report_call(f, (Array{T,3} where {T<:Real},))

# works:
report_call(f, (Array{Float64,3},)) 
```

You can probably ignore many of my review comments as they are mostly about me removing comments lol

harrisonritz:
I need to double-check with `Julia +release` so the JET tests actually run

rsenne:
hrm weird this approach worked locally....let me see if i can debug

rsenne:
Huzzah!

rsenne (approved):

PR #126, opened 2026-07-04
Feature/null models

### LINKED ISSUE #125: Add null models
Compare LDS to observation-level null models:

- intercept only
- inputs only
- vector auto-regression (VAR)
- VAR + inputs

Compute train and test loglik for null models, and compute relative Cox-Snell R^2 against LDS.

harrisonritz:
Adds latent-free baselines to score a fitted model against, and the StatsAPI
model-comparison surface built on them. Resolves #125.

## Baselines

`AffineNullModel{T}` is one affine-Gaussian baseline whose two structural
switches (`lag`, `input_shift`) are `const` fields, so the four named variants
are configurations of one type rather than four implementations:

| baseline | model |
|---|---|
| `:intercept` | `y_t ~ N(d, R)` |
| `:inputs` | `y_t ~ N(d + D v_{t-shift}, R)` |
| `:var` | `y_1 ~ N(μ₀, R₀)`; `y_t ~ N(F y_{t-1} + d, R)` |
| `:var_inputs` | `y_1 ~ N(μ₀, R₀)`; `y_t ~ N(F y_{t-1} + d + D v_{t-shift}, R)` |

`input_shift` matches the LDS input timing: `1` for dynamics inputs (`ux`, which
act on `x_{t-1}`), `0` for observation inputs (`uy`). Fit with
`fit!(null, y; inputs)` and score with `loglikelihood(null, y; inputs)` — the
plug-in log-likelihood, so passing held-out data gives a held-out score. An
optional `MNPrior` on the stacked matrix `[F d D]` and `IWPrior`s on `R` / `R₀`
mirror the LDS M-step.

One design-matrix builder, one regression kernel and one scoring kernel serve
all four: `:intercept` / `:var` are the zero-column-`D` special cases of
`:inputs` / `:var_inputs` by construction. Degenerate inputs are inert rather
than fatal — an input baseline on a channel the LDS does not use collapses to
its input-free counterpart.

## Model comparison on a fitted LDS

- `r2(lds, y, variant; null=:intercept, null_inputs=:ux)` — all three StatsBase
  pseudo-R² variants (`:CoxSnell` default, `:McFadden`, `:Nagelkerke`),
  comparing the LDS **marginal** log-likelihood (latents integrated out) to the
  baseline's plug-in log-likelihood on the same data.
- `nullloglikelihood(lds, y)` — the intercept baseline's plug-in log-likelihood.
- `nobs(lds, y)` — `obs_dim * sum(tsteps)`, the scale both log-likelihoods live
  on.

The baseline inherits the LDS's `R_prior` by default, so both estimate `R` under
the same IW prior. Gaussian LDS only: a Poisson LDS has no tractable marginal
and an `SLDS` is a different type, so both raise a `MethodError` by design.

## Errors

A rank-deficient design throws a `NumericalStabilityError` naming the cause
rather than a bare `SingularException`. The case that reaches users is an input
block carrying its own constant row, which duplicates the intercept every
baseline fits.

## Tests

30 test functions in `test/NullModels/NullModels.jl` (142 assertions), checking
against independent closed-form references rather than the implementation's own
kernels: the analytic intercept MLE and Gaussian log-density, VAR(1) parameter
recovery, held-out plug-in scoring, capacity ordering across the four baselines,
input-shift alignment, ragged and single-trial data, the IW/MN prior log-terms,
the `r2` variant formulas, and the error paths.


harrisonritz:
@rsenne ready

src/lds/fit_LDS.jl
```diff
@@ -1262,3 +1262,128 @@ function loglikelihood(
...
+            SiPCt .= PCt'
+            ldiv!(S_ch, SiPCt)
+            mul!(P_f, PCt, SiPCt)
+            P_f .= P_p .- P_f
+            Symmetrize!(P_f)
+        end
+    end
+
+    return total_ll
+end
```
rsenne, line 1110: Would it be possible to make thi spart of the existing function? I.e., where there is no input it just calculates the regular marginal? 

This is also related to a thought I had where many expressions are currently repeated across the codebase and thus suggest we should have some smaller functions peppered around e.g., `P_p - PCt S⁻¹ PCt'`something like this should be able to be generalized.

Thats the subject for another PR though.

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+    )
+
+    return (
+        intercept=intercept_res, inputs=inputs_res, var=var_res, var_inputs=var_inputs_res
+    )
+end
+
+# -----------------------------------------------------------------------------
+# Input validation
+# -----------------------------------------------------------------------------
```
rsenne: Can we do
```suggestion
#=
Input validation
=#
```

or

```suggestion
# Input validation
```

please?

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+                size(v),
+            ),
+        )
+    end
+    return nothing
+end
+
+# -----------------------------------------------------------------------------
+# Core regression + log-likelihood helpers
+# -----------------------------------------------------------------------------
```
rsenne: Just flagging see above

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+    Symmetrize!(EE)
+    ldiv!(F.U', EE)
+    ldiv!(F.U, EE)
+
+    return T(-0.5) * (T(n) * (T(obs_dim) * log(T(2π)) + log_det_R) + tr(EE))
+end
+
+# -----------------------------------------------------------------------------
+# Data stacking helpers
+# -----------------------------------------------------------------------------
```
rsenne: flagging

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+
+# Stack y_1 across trials → (obs_dim × ntrials).
+@inline function _stack_y_init(y::AbstractArray{T,3}) where {T<:Real}
+    obs_dim = size(y, 1)
+    return reshape(y[:, 1, :], obs_dim, size(y, 3))
+end
+
+# -----------------------------------------------------------------------------
+# Model 1: intercept only
+# -----------------------------------------------------------------------------
```
rsenne: ibid

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+    end
+
+    d = vec(W[:, 1])
+    params = (d=d, R=R)
+    return (train_ll=train_ll, test_ll=test_ll, params=params)
+end
+
+# -----------------------------------------------------------------------------
+# Model 2: inputs only (no autocorrelation)
+# -----------------------------------------------------------------------------
```
rsenne: ibid

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+
+    d = vec(W[:, 1])
+    D = v_dim > 0 ? W[:, 2:end] : Matrix{T}(undef, size(W, 1), 0)
+    params = (d=d, D=D, R=R)
+    return (train_ll=train_ll, test_ll=test_ll, params=params)
+end
+
+# -----------------------------------------------------------------------------
+# Model 3: VAR(1) only (no inputs)
+# -----------------------------------------------------------------------------
```
rsenne: ibid

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+    μ_0 = vec(W_init[:, 1])
+    F = W_var[:, 1:obs_dim]
+    d = vec(W_var[:, obs_dim + 1])
+    params = (μ_0=μ_0, R_0=R0, F=F, d=d, R=R)
+    return (train_ll=train_ll, test_ll=test_ll, params=params)
+end
+
+# -----------------------------------------------------------------------------
+# Model 4: VAR(1) + inputs
+# -----------------------------------------------------------------------------
```
rsenne: ibid

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+# LDS marginal likelihood. Priors therefore enter only when *fitting* the null
+# parameters: the LDS's observation-noise prior `obs_model.R_prior` is forwarded
+# to the null models' `R_prior` so both estimate `R` under the same IW prior.
+# (The LDS's remaining priors — `Q`, `P0`, `[A B]`, `[C D]` — live in latent
+# space and have no null-model counterpart, so they are not transferred.)
+#
+# Restricted to the Gaussian LDS: a Poisson LDS (`PoissonObservationModel`) has
+# no tractable marginal likelihood, and an `SLDS` is a different type entirely,
+# so both fall through to a `MethodError` by design.
+# =============================================================================
```
rsenne:
```suggestion
#=
Cox–Snell (likelihood-ratio) R² of an LDS versus each null baseline.

`compute_R2` scores a *fitted* Gaussian LDS against the four `test_null`
baselines using the Cox–Snell R²

R² = 1 - exp( (2/n) · (loglik_null - loglik_model) )

where `n = obs_dim · tsteps · ntrials` is the total scalar-observation count.
`loglik_model` is the LDS **marginal** (observed-data) log-likelihood with
latent states integrated out — the same quantity as `loglikelihood(lds, y)`
but extended here to honour the LDS's `B·ux` / `D·uy` input terms, which the
base `loglikelihood` drops. `loglik_null` is the plug-in Gaussian data
log-density of a null model fit on the training split — the `test_null`
`test_ll` convention (no prior terms), which is the null-side analogue of the
LDS marginal likelihood. Priors therefore enter only when *fitting* the null
parameters: the LDS's observation-noise prior `obs_model.R_prior` is forwarded
to the null models' `R_prior` so both estimate `R` under the same IW prior.
(The LDS's remaining priors — `Q`, `P0`, `[A B]`, `[C D]` — live in latent
space and have no null-model counterpart, so they are not transferred.)

Restricted to the Gaussian LDS: a Poisson LDS (`PoissonObservationModel`) has
no tractable marginal likelihood, and an `SLDS` is a different type entirely,
so both fall through to a `MethodError` by design.
=#
```

rsenne: Can we make this less verbose?

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+# Cox–Snell R² from a pair of total log-likelihoods and the observation count.
+@inline function _cox_snell_r2(ll_null::T, ll_model::T, n::Int) where {T<:Real}
+    return one(T) - exp((T(2) / T(n)) * (ll_null - ll_model))
+end
+
+# Fit the four null baselines on `train_data` and return their plug-in Gaussian
+# data log-densities evaluated on `eval_data` (the `test_null` `test_ll`
+# convention, no prior terms). `null_inputs` selects which of the LDS's input
+# channels the input-bearing baselines consume: `"ux"` uses `data.ux` (the
+# `test_null` default), `"uy"` uses `data.uy`.
```
rsenne: This exact phrasing is very close to below--can we jsut update the comment below to reflect whats different? Also--can we use the idiomatic block quote syntax?

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+- `R_prior::Union{Nothing,IWPrior{T}} = lds.obs_model.R_prior`: IW prior used
+  when fitting each baseline's observation covariance `R`, defaulting to the
+  LDS's own observation-noise prior so both are regularized alike.
+
+# Returns
+A `NamedTuple` keyed by baseline name. For the single-`data` form each entry is
+`(R2, lds_ll, null_ll, n)`; for the `train_data`/`test_data` form each entry is
+`(train_R2, test_R2, lds_train_ll, lds_test_ll, null_train_ll, null_test_ll,
+n_train, n_test)`.
+"""
```
rsenne: Can we make this a bit less verbose?

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+    R0_prior::Union{Nothing,IWPrior{T}},
+) where {T<:Real}
+    obs_dim, tsteps, _ = size(train_data.y)
+    tsteps >= 2 ||
+        throw(ArgumentError("VAR(1) null model requires tsteps ≥ 2 (got $tsteps)"))
+
+    # Init regression: y_1[:, n] ~ N(μ_0, R_0); regress on a constant.
+    Y_init = _stack_y_init(train_data.y)
+    X_init = _bias_row(T, size(Y_init, 2))
+    W_init, R0, S_init, n_init = _null_fit_regression(Y_init, X_init, nothing, R0_prior)
```
rsenne: This will throw a `PosDef` exception on single trial data

test/NullModels/NullModels.jl
```diff
@@ -0,0 +1,602 @@
...
+function test_null_test_ll_nothing_when_no_test_data(rng=MersenneTwister(0xC0FFEE))
+    data = _null_make_data(rng, 2, 10, 3)
+    res = test_null(data)
+    @test res.intercept.test_ll === nothing
+    @test res.inputs.test_ll === nothing
+    @test res.var.test_ll === nothing
+    @test res.var_inputs.test_ll === nothing
+end
+
+# ----- Collapse identities --------------------------------------------------
```
rsenne:
```suggestion
# Collapse identities
```

test/NullModels/NullModels.jl
```diff
@@ -0,0 +1,602 @@
...
+
+    res = test_null(data)
+
+    # With v_dim = 0 the inputs model has the same regressors as the intercept
+    # model, and the VAR+inputs model matches the VAR-only model.
+    @test res.inputs.train_ll ≈ res.intercept.train_ll atol = 1e-10
+    @test res.var_inputs.train_ll ≈ res.var.train_ll atol = 1e-10
+end
+
+# ----- Signal recovery + LL ordering ---------------------------------------
```
rsenne:
```suggestion
# Signal recovery + LL ordering
```

test/NullModels/NullModels.jl
```diff
@@ -0,0 +1,602 @@
...
+    @test res.var_inputs.train_ll > res.var.train_ll
+    @test res.inputs.params.D ≈ D_true rtol = 0.15
+    @test res.inputs.params.d ≈ d_true rtol = 0.2
+    # Shape sanity on returned parameters.
+    @test size(res.inputs.params.D) == (obs_dim, v_dim)
+    @test size(res.var_inputs.params.D) == (obs_dim, v_dim)
+    @test size(res.var_inputs.params.F) == (obs_dim, obs_dim)
+end
+
+# ----- VAR(1) parameter recovery -------------------------------------------
```
rsenne:
```suggestion
# VAR(1) parameter recovery
```

test/NullModels/NullModels.jl
```diff
@@ -0,0 +1,602 @@
...
+    @test res.var.params.F ≈ F_true atol = 0.05
+    @test res.var.params.R ≈ R_true atol = 0.02
+    @test length(res.var.params.μ_0) == obs_dim
+    @test size(res.var.params.R_0) == (obs_dim, obs_dim)
+    @test length(res.var.params.d) == obs_dim
+    # VAR model should beat intercept-only on truly autocorrelated data.
+    @test res.var.train_ll > res.intercept.train_ll
+end
+
+# ----- IW-prior LL shift identity ------------------------------------------
```
rsenne:
```suggestion
# IW-prior LL shift identity
```

test/NullModels/NullModels.jl
```diff
@@ -0,0 +1,602 @@
...
+    @test res.inputs.train_ll - data_ll_ref ≈ expected_shift atol = 1e-8
+
+    # Strong shrinkage should also observably pull the MAP toward M₀ = 0.
+    strong_prior = MNPrior(; M₀=M₀, Λ=Matrix{T}(1e6 * I, 1 + v_dim, 1 + v_dim))
+    res_strong = test_null(data; inputs_W_prior=strong_prior)
+    W_strong = hcat(res_strong.inputs.params.d, res_strong.inputs.params.D)
+    @test norm(W_strong) < norm(W)
+end
+
+# ----- Input override + validation -----------------------------------------
```
rsenne:
```suggestion
# Input override + validation
```

test/NullModels/NullModels.jl
```diff
@@ -0,0 +1,602 @@
...
+    # When test_inputs is not supplied, defaults to test_data.ux — differs
+    # from what happens if we explicitly pass zeros.
+    train = _null_make_data(rng, 2, 12, 4; v_dim=1)
+    test = _null_make_data(rng, 2, 10, 3; v_dim=1)
+    res = test_null(train; test_data=test)
+    @test res.inputs.test_ll isa Float64 && isfinite(res.inputs.test_ll)
+    @test res.var_inputs.test_ll isa Float64 && isfinite(res.var_inputs.test_ll)
+end
+
+# ----- Error paths ---------------------------------------------------------
```
rsenne:
```suggestion
# Error paths
```

test/NullModels/NullModels.jl
```diff
@@ -0,0 +1,602 @@
...
+    data = _null_make_var_data(rng, F_true, d_true, R_true, μ_0, R_0, tsteps, ntrials)
+    R0_prior = IWPrior(; Ψ=Matrix{T}(0.1 * I, obs_dim, obs_dim), ν=T(obs_dim + 2))
+    res = test_null(data; R0_prior=R0_prior)
+
+    @test res.var.train_ll > res.intercept.train_ll
+    @test res.var_inputs.train_ll ≈ res.var.train_ll atol = 1e-10
+    @test res.inputs.train_ll ≈ res.intercept.train_ll atol = 1e-10
+end
+
+# ----- All-priors path finite ---------------------------------------------
```
rsenne:
```suggestion
# All-priors path finite
```

test/NullModels/NullModels.jl
```diff
@@ -0,0 +1,602 @@
...
+    for name in (:intercept, :inputs, :var, :var_inputs)
+        entry = getproperty(res, name)
+        @test isfinite(entry.train_ll)
+        @test isfinite(entry.test_ll)
+    end
+end
+
+# ===========================================================================
+# Cox–Snell R² (`compute_R2`) against the null baselines.
+# ===========================================================================
```
rsenne:
```suggestion
#=
Cox–Snell R² (`compute_R2`) against the null baselines.
=#
```

src/lds/fit_LDS.jl
```diff
@@ -1262,3 +1262,128 @@ function loglikelihood(
...
+"""
+function loglikelihood(
+    lds::LinearDynamicalSystem{T,SM,OM}, data::Data{T}
+) where {T<:Real,SM<:GaussianStateModel{T},OM<:GaussianObservationModel{T}}
+    A = lds.state_model.A
+    Q = lds.state_model.Q
+    b = lds.state_model.b
+    x0 = lds.state_model.x0
+    P0 = lds.state_model.P0
+    B = lds.state_model.B
```
rsenne: What happens if someone passes in a degenerate matrix? E.g., lets say we fit a null model where its just an input regression and so B has zero columns and ux is irrelevant? I think this case would throw?

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+# i.e. plug-in MAP parameters with the same partial-constant prior
+# contributions that the SSM ELBO uses. The test LL is the plug-in Gaussian
+# log-density on the test arrays (no prior terms), matching the convention of
+# the Kalman-filter marginal `loglikelihood(lds, y)`.
+#
+# The four baselines are:
+#   1. intercept   y_t ~ N(d, R)
+#   2. inputs      y_t ~ N(d + D v_t, R)
+#   3. var         y_1 ~ N(μ_0, R_0);  y_t ~ N(F y_{t-1} + d, R)         (t ≥ 2)
+#   4. var_inputs  y_1 ~ N(μ_0, R_0);  y_t ~ N(F y_{t-1} + d + D v_t, R) (t ≥ 2)
```
rsenne: Another potetial null is PPCA

rsenne: not neccesary just a thought

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+
+# Fit the four null baselines on `train_data` and return their plug-in Gaussian
+# data log-densities evaluated on `eval_data` (the `test_null` `test_ll`
+# convention, no prior terms). `null_inputs` selects which of the LDS's input
+# channels the input-bearing baselines consume: `"ux"` uses `data.ux` (the
+# `test_null` default), `"uy"` uses `data.uy`.
+function _null_plugin_lls(
+    train_data::Data{T},
+    eval_data::Data{T},
+    null_inputs::AbstractString,
```
rsenne: I think Julia prefers `Symbol` types over strings

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+
+# Fit the four null baselines once on `train_data`, then return their plug-in
+# Gaussian data log-densities on BOTH `train_data` and `test_data` as
+# `(train_lls, test_lls)`.  A single `test_null` call handles the fitting and
+# scores `test_data` as a by-product; train plug-in LLs are then computed by
+# re-evaluating `train_data` against the already-fitted parameters.
+function _null_plugin_lls_both(
+    train_data::Data{T},
+    test_data::Data{T},
+    null_inputs::AbstractString,
```
rsenne: ibid

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+# Returns
+A `NamedTuple` keyed by baseline name. For the single-`data` form each entry is
+`(R2, lds_ll, null_ll, n)`; for the `train_data`/`test_data` form each entry is
+`(train_R2, test_R2, lds_train_ll, lds_test_ll, null_train_ll, null_test_ll,
+n_train, n_test)`.
+"""
+function compute_R2(
+    lds::LinearDynamicalSystem{T,SM,OM},
+    data::Data{T};
+    null_inputs::AbstractString="ux",
```
rsenne: ibid

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+    return map(null_ll) do ll
+        return (R2=_cox_snell_r2(ll, lds_ll, n), lds_ll=lds_ll, null_ll=ll, n=n)
+    end
+end
+
+function compute_R2(
+    lds::LinearDynamicalSystem{T,SM,OM},
+    train_data::Data{T},
+    test_data::Data{T};
+    null_inputs::AbstractString="ux",
```
rsenne: ibid

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+intractable / handled elsewhere).
+
+# Arguments
+- `lds::LinearDynamicalSystem`: a fitted Gaussian LDS.
+- `data::Data{T}`: single dataset — returns one R² per baseline.
+- `train_data, test_data::Data{T}`: fit the baselines on `train_data` and score
+  both splits — returns `train_R2` and `test_R2` per baseline.
+
+# Keyword Arguments
+- `null_inputs::AbstractString = "ux"`: which LDS input channel the
```
rsenne: ibid

src/lds/fit_LDS.jl
rsenne: also all of our loglikelihoods should be overloads of StatsAPI would you mind adding that to all relevant functions?

src/lds/fit_LDS.jl
```diff
@@ -1262,3 +1262,128 @@ function loglikelihood(
...
+            end
+
+            # Innovation covariance: S = C P_p C' + R.
+            mul!(PCt, P_p, C')
+            mul!(Smat, C, PCt)
+            Smat .+= R
+            Symmetrize!(Smat)
+
+            # One-step predictive log-likelihood: log N(e; 0, S).
+            S_ch = cholesky(Hermitian(Smat))
```
rsenne: This could be a `cholesky!` (applies to above as well but, i think we can refactor these two functions into one)

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+# `(train_lls, test_lls)`.  A single `test_null` call handles the fitting and
+# scores `test_data` as a by-product; train plug-in LLs are then computed by
+# re-evaluating `train_data` against the already-fitted parameters.
+function _null_plugin_lls_both(
+    train_data::Data{T},
+    test_data::Data{T},
+    null_inputs::AbstractString,
+    R_prior::Union{Nothing,IWPrior{T}},
+) where {T<:Real}
+    v_train, v_test = if null_inputs == "ux"
```
rsenne: Should the null input be ux_t or ux_{t-1} to match the LDS input scheme? IIUC we do the former when the latter would match more faithfully

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+    init_vi_tr = _null_test_ll(Y_init_tr, _bias_row(T, n_init_tr), W_init_vi, p_vi.R_0)
+    W_var_vi = hcat(p_vi.F, reshape(p_vi.d, obs_dim, 1), p_vi.D)
+    X_var_vi_tr = vcat(
+        _stack_y_prev(train_data.y), _bias_row(T, n_var_tr), _stack_inputs_next(v_train)
+    )
+    var_inputs_tr = init_vi_tr + _null_test_ll(Y_next_tr, X_var_vi_tr, W_var_vi, p_vi.R)
+
+    train_lls = (
+        intercept=intercept_tr, inputs=inputs_tr, var=var_tr, var_inputs=var_inputs_tr
+    )
```
rsenne: doesn't `_null_trial_ll` compute this already?

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+    else
+        Y_te = _stack_y_all(test_data.y)
+        X_te = _bias_row(T, size(Y_te, 2))
+        _null_test_ll(Y_te, X_te, W, R)
+    end
+
+    d = vec(W[:, 1])
+    params = (d=d, R=R)
+    return (train_ll=train_ll, test_ll=test_ll, params=params)
+end
```
rsenne: This is just a special case of the `_null_inputs` can we delete this?

src/stats/null_models.jl
rsenne: Simialr to above we should be able to delete this and lean on `null_var_inputs`

src/stats/null_models.jl
```diff
@@ -0,0 +1,691 @@
...
+        throw(
+            DimensionMismatchError(
+                "$name shape (input_dim, T, ntrials)",
+                (size(v, 1), tsteps, ntrials),
+                size(v),
+            ),
+        )
+    end
+    return nothing
+end
```
rsenne: isn't this the same as `_check_inputs` in `fit_LDS`?

rsenne (changes requested):

rsenne:
Okay sorry for the long review--but I have more to say here too lol

I think we should redesign the null models to lean on some structs and use standard StatsAPI methods. I think we could do this all in one big PR or we can break this PR down into a few smaller ones. Your call on that! Here is my design thought.

Right now the proposed API is quite similar to how many packages have null models already so i think we should redesign around StatsAPI functions to make this more in line with the current ecosystem.

Concreteky:

1. Make each baseline a small struct (`InterceptNull`, `VARNull`, ...) with `fit!(null, data)` and `loglikelihood(null, data)` methods.
2. Extend `StatsAPI.r2(lds, data; null=:intercept)` instead of exporting `compute_R2`. 
3. `StatsAPI.nullloglikelihood(lds, data)` falls out for free (it's the intercept baseline's plug-in LL)
4. We should also be able to implement `StatsAPI.nobs`

Also--sorry for being such a hard ass about comments etc. just trying to keep the style uniform as possible (i have intentions of reformatting many of the current code comments soon)

LMK how all of this sounds!

harrisonritz:
Don't apologize, this is super helpful. I'll take a pass later this week.

rsenne:
i'm going to rebase/squash the branch because the diff history is corrupted

rsenne:
i'll also be reviewing your other pr today sorry for the delay!

harrisonritz:
Shit, sorry about the got history.

Not urgent, I'm "on vacation", but appreciate whenever you can take a look at it. 

Lmk if it's too sloppy and I'll work through it before your full review.

harrisonritz:
See how it looks after the rebase -- if it still has way too many irrelevant changes, lmk and I'll just make another PR.

rsenne:
@harrisonritz you'll need to run

```
git fetch origin
git checkout feature/null-models
git reset --hard origin/feature/null-models
```

(or upstream if not origin) before pullin/pushing

harrisonritz:
I'll do that for next push. FWIW I think this looks reviewable as is (but no rush)

src/utils/show.jl
```diff
@@ -68,13 +68,13 @@ function Base.show(io::IO, gom::GaussianObservationModel; gap="")
...
-    nobs, nstate = size(pom.C)
+    obs_dim, nstate = size(pom.C)
 
     println(io, gap, "Poisson Observation Model:")
     println(io, gap, "--------------------------")
 
-    if nobs > 4 || nstate > 4
-        println(io, gap, " size(C) = ($nobs, $nstate)")
+    if obs_dim > 4 || nstate > 4
+        println(io, gap, " size(C) = ($obs_dim, $nstate)")
```
rsenne, line 77: nice

src/stats/null_models.jl
```diff
@@ -0,0 +1,543 @@
...
+        Matrix{T}(I, lag_dim, lag_dim),
+        W_prior,
+        R_prior,
+        R₀_prior,
+    )
+end
+
+function AffineNullModel{T}(
+    baseline::Symbol, obs_dim::Int; input_dim::Int=0, kwargs...
+) where {T<:Real}
```
rsenne, line 111: fdor this covnenience wrapper--we need some logic preventing someone from doing something dumb by mistake. Right now, a user can pass `lag` in the `kwargs`. In Julia the later var wins, so if someone did, say:

```julia
null = AffineNullModel{Float64}(:intercept, 3; lag=true)
size(null.F) # =(3 , 3)
```

so basically they'll get a VAR despite requesting an intercept

src/stats/null_models.jl
```diff
@@ -0,0 +1,543 @@
...
+        if null.input_dim > 0
+            push!(blocks, _shifted_inputs(vn, Val(SHIFT))[:, rows])
+        end
+
+        push!(responses, yn[:, rows])
+        push!(designs, reduce(vcat, blocks))
+    end
+
+    isempty(responses) &&
+        throw(ArgumentError("a VAR baseline needs at least one trial with tsteps ≥ 2"))
```
rsenne, line 229: this could throw on an intercept only model so we need some more control flow here

src/stats/null_models.jl
```diff
@@ -0,0 +1,543 @@
...
+
+The `μ₀`/`R₀` step of the VAR baselines mirrors the LDS's `(x0, P0)` layer, so
+every timestep is scored and the baseline stays comparable to the LDS marginal
+likelihood term for term.
+
+Baselines are fit with `fit!` and scored with `loglikelihood`; `r2`,
+`nullloglikelihood` and `nobs` build on those to score a fitted LDS against a
+baseline. Priors are optional and mirror the LDS M-step: an `MNPrior` on the
+stacked regression matrix, `IWPrior`s on `R` and `R₀`.
+=============================================================================#
```
rsenne:
```suggestion
#=============================================================================
Latent-free affine-Gaussian baselines for LDS comparison.

`AffineNullModel{LAG,SHIFT}` covers four configurations:

  :intercept   y_t ~ N(d, R)
  :inputs      y_t ~ N(d + D v_{t-SHIFT}, R)
  :var         y_1 ~ N(μ₀, R₀); y_t ~ N(F y_{t-1} + d, R)
  :var_inputs  y_1 ~ N(μ₀, R₀); y_t ~ N(F y_{t-1} + d + D v_{t-SHIFT}, R)

Here `LAG` enables the autoregressive term and `SHIFT` matches the corresponding
LDS input timing: lagged for dynamics inputs and contemporaneous for observation
inputs. Pretrial inputs are zero. For VAR models, `(μ₀, R₀)` scores the first
observation analogously to the LDS initial-state layer.

Use `fit!` and `loglikelihood`; `r2`, `nullloglikelihood`, and `nobs` compare a
fitted LDS against a baseline. Optional priors are an `MNPrior` on the stacked
regression matrix and `IWPrior`s on `R` and `R₀`.
=============================================================================#
```

rsenne: this read a little too claude to me

src/stats/null_models.jl
```diff
@@ -0,0 +1,543 @@
...
+
+```julia
+null = AffineNullModel{Float64}(:intercept, obs_dim)
+fit!(null, y_train)
+ll  = loglikelihood(lds, y_test)
+ll₀ = loglikelihood(null, y_test)
+r2_heldout = 1 - exp(2 * (ll₀ - ll) / nobs(lds, y_test))
+```
+
+See also [`nullloglikelihood`](@ref), [`nobs`](@ref), [`AffineNullModel`](@ref).
```
rsenne: I don't add examples. I'm not a fan of the doctest interface in Julia and without that we are liable for these to go stale. Can you remove?

src/stats/null_models.jl
```diff
@@ -0,0 +1,543 @@
...
+- `input_dim::Int`: input dimension (`0` when the baseline takes no inputs).
+- `d::Vector{T}`: intercept.
+- `D::Matrix{T}`: input matrix (`obs_dim × input_dim`).
+- `F::Matrix{T}`: autoregressive matrix (`obs_dim × obs_dim`, empty when `!LAG`).
+- `R::Matrix{T}`: innovation covariance.
+- `μ₀::Vector{T}`, `R₀::Matrix{T}`: initial-step mean/covariance (empty when `!LAG`).
+- `W_prior`, `R_prior`, `R₀_prior`: optional MN/IW priors, mirroring the LDS M-step.
+
+See also [`r2`](@ref), [`nullloglikelihood`](@ref).
+"""
```
rsenne, line 25:
```suggestion
"""
    AffineNullModel{T,LAG,SHIFT}

Affine-Gaussian null model for a fitted LinearDynamicalSystem. LAG
(Bool) controls whether the model includes the autoregressive term
F y_{t-1}. SHIFT (0 or 1) gives the input lag. When the model has no
inputs, D has zero columns, consistent with the LDS representation.

# Constructors
    AffineNullModel{T}(obs_dim; input_dim=0, lag=false, input_shift=0, priors...)
    AffineNullModel{T}(baseline::Symbol, obs_dim; input_dim=0, input_shift=0, priors...)

The second form selects one of $(NULL_BASELINES) and sets lag and
input_dim. For the input-free baselines :intercept and :var, it sets
input_dim=0. Set input_shift=0 to score v_t (the LDS observation input) or
input_shift=1 to score v_{t-1} (the LDS dynamics input).

# Fields
- `obs_dim::Int`: observation dimension.
- `input_dim::Int`: input dimension (`0` when the baseline takes no inputs).
- `d::Vector{T}`: intercept.
- `D::Matrix{T}`: input matrix (`obs_dim × input_dim`).
- `F::Matrix{T}`: autoregressive matrix (`obs_dim × obs_dim`, empty when `!LAG`).
- `R::Matrix{T}`: innovation covariance.
- `μ₀::Vector{T}`, `R₀::Matrix{T}`: initial-step mean/covariance (empty when `!LAG`).
- `W_prior`, `R_prior`, `R₀_prior`: optional MN/IW priors, mirroring the LDS M-step.

See also [`r2`](@ref), [`nullloglikelihood`](@ref).
"""
```

src/stats/null_models.jl
```diff
@@ -0,0 +1,543 @@
...
+    )
+end
+
+#=
+Trial normalization. A baseline is a standalone model, so it cannot reuse the
+`Data(lds, y; ux, uy)` constructor — that validates against an LDS. It accepts
+the same public shape family instead and canonicalizes here: a single
+`(obs_dim, T)` matrix, an `(obs_dim, T, ntrials)` array, or a vector of per-trial
+matrices (ragged lengths allowed).
+=#
```
rsenne:
```suggestion
#=
Normalize the supported trial formats: an (obs_dim, T) matrix, an
(obs_dim, T, ntrials) array, or a vector of per-trial matrices (which may have
different lengths). The baseline cannot use Data(lds, y; ux, uy), because that
constructor validates the data against an LDS.
=#
```

rsenne: we could also add a `Data` variant ofr the NullModels too if desired

src/stats/null_models.jl
```diff
@@ -0,0 +1,543 @@
...
+    return null
+end
+
+"""
+    loglikelihood(null::AffineNullModel, y; inputs=nothing)
+
+Plug-in log-likelihood of `y` under the fitted baseline — no prior terms, so it
+is directly comparable to the LDS marginal `loglikelihood(lds, y)`. Scoring data
+other than the fit data gives the held-out log-likelihood.
+"""
```
rsenne, line 389:
```suggestion
"""
loglikelihood(null::AffineNullModel, y; inputs=nothing)

Plug-in log-likelihood of `y` under the fitted baseline. This excludes prior
terms and is directly comparable to the LDS marginal `loglikelihood(lds, y)`.
Pass held-out data to compute held-out log-likelihood.
"""
```

src/stats/null_models.jl
```diff
@@ -0,0 +1,543 @@
...
+    return loglikelihood(null, data.y, inputs)
+end
+
+"""
+    nobs(lds::LinearDynamicalSystem, y; ux=nothing, uy=nothing)
+
+Number of scalar observations in `y` — `obs_dim * sum(tsteps)`, summed over
+trials. This is the denominator [`r2`](@ref) uses for the Cox–Snell and
+Nagelkerke variants, so it is on the same scale as `loglikelihood(lds, y)`.
+"""
```
rsenne, line 471:
```suggestion
"""
nobs(lds::LinearDynamicalSystem, y; ux=nothing, uy=nothing)

Number of scalar observations in `y`: `obs_dim * sum(tsteps)` across trials.
r2 uses this count in the Cox–Snell and Nagelkerke variants.
"""
```

src/stats/null_models.jl
```diff
@@ -0,0 +1,543 @@
...
+
+"""
+    nullloglikelihood(lds, y; ux=nothing, uy=nothing, R_prior=lds.obs_model.R_prior)
+
+Plug-in log-likelihood of the intercept baseline `y_t ~ N(d, R)` fit on the same
+data — the reference point [`r2`](@ref) measures `lds` against by default.
+
+`R_prior` defaults to the LDS's own observation-noise prior so the baseline's `R`
+is regularized exactly as the LDS's is.
+"""
```
rsenne, line 487:
```suggestion
"""
nullloglikelihood(lds, y; ux=nothing, uy=nothing, R_prior=lds.obs_model.R_prior)

Plug-in log-likelihood of the intercept baseline `y_t ~ N(d, R)`, fit to the
same data. This is the default reference model for `r2`.

`R_prior` defaults to the LDS observation-noise prior.
"""
```

src/stats/null_models.jl
```diff
@@ -0,0 +1,543 @@
...
+- `d::Vector{T}`: intercept.
+- `D::Matrix{T}`: input matrix (`obs_dim × input_dim`).
+- `F::Matrix{T}`: autoregressive matrix (`obs_dim × obs_dim`, empty when `!LAG`).
+- `R::Matrix{T}`: innovation covariance.
+- `μ₀::Vector{T}`, `R₀::Matrix{T}`: initial-step mean/covariance (empty when `!LAG`).
+- `W_prior`, `R_prior`, `R₀_prior`: optional MN/IW priors, mirroring the LDS M-step.
+
+See also [`r2`](@ref), [`nullloglikelihood`](@ref).
+"""
+mutable struct AffineNullModel{T<:Real,LAG,SHIFT}
```
rsenne: I'm not sure this needs three type params. These models in theory have their function run once so not much for the compiler to specialzie on. Soemthing like:

`if null.lag` control flow would have same effect and no measurable performance hit imo

src/stats/null_models.jl
```diff
@@ -0,0 +1,543 @@
...
+    loglikelihood(null::AffineNullModel, y; inputs=nothing)
+
+Plug-in log-likelihood of `y` under the fitted baseline — no prior terms, so it
+is directly comparable to the LDS marginal `loglikelihood(lds, y)`. Scoring data
+other than the fit data gives the held-out log-likelihood.
+"""
+function StatsAPI.loglikelihood(null::AffineNullModel{T}, y; inputs=nothing) where {T<:Real}
+    trials = _null_trials(y)
+    tsteps = Int[size(yn, 2) for yn in trials]
+    v = _null_input_trials(T, inputs, tsteps, null.input_dim)
```
rsenne: nit but these three lines are copied verbatim in `fit!` and `_null_logmap` could make this a helper func

rsenne (changes requested):
overall, loogs, good. A few nits and suggestions. Mostly just suggestions to de-claude speak this. in general i don't like how claude documents things. Has a tendency to bloat the code and also imo it can be somewhat hard to read

rsenne:
overall looks good. Some nits and suggestions. Biggest thing is reducing some claude-speak. I tend to find its docs hard to read so i'm on the side of using it less for that

src/stats/null_models.jl
```diff
@@ -0,0 +1,630 @@
...
+    obs_dim = size(first(data.y), 1)
+
+    # The input-free baselines reject an input block, so build them without one
+    # and hand them matching zero-row inputs.
+    null, inputs = if _null_has_inputs(baseline)
+        AffineNullModel{T}(
+            baseline,
+            obs_dim;
+            input_dim=size(first(v), 1),
+            input_shift=shift,
```
rsenne: `input_shift = size(first(v),1) > 0 ? shift : 0`

this is needed to prevent a degenerate case

rsenne: we also need to add tests where input_dim=0

rsenne:
Can we also go through again de-claude further? i still think there is a good bit of verbosity here (i do think codex can be good at fixing this if you have access)

rsenne (approved):

Issue #131, opened 2026-07-07
Remove duplicate code


rsenne:
Right now there is considerable code dupe for functions that are repeated across the Gaussian/Poisson LDS models. E.g., in the joints we need to compute the joints which means we calculate $N(Ax_{t-1} + B u_t + b, Q)$ these computations can be generalized into something like `stateloglikelihood!` or something that we can dispatch based on what time-point it is. Same can be said for obs likelihood etc.

This has the advantage of 1.) reducing code duplication and 2.) if we make a change to these functions we update it in one place and not several.

rsenne:
- [x] Deduplicate code in loglikelihoods
- [x] Deduplicate code in gradients
- [x] Deduplicate code in hessians

rsenne:
Completed in #135 / #136 / #141

PR #135, opened 2026-07-07
Fix/131 dedupe lds likelihoods

rsenne:
PR to partially address #131. 

Split the `joint_loglikelihood` into a `stateloglikelihood!` and an `obsloglikelihood!` (not deadset on names). The rationale for this is 1.) we don't re-implement the same calculation a bunch of places (prior to this we had the same logic in three places and one of them was wrong---SLDS 2.) to avoid differences between models e.g., my previous comment. 3.) to eventually make it s.t. we can offer a fallback `joint_loglikelihood` for generic models, which I'd like to make it easier to do

- Adds tests to check against Distributions
- Fixes the SLDS bug
- Adds stateloglikelihood!` and `obsloglikelihood!` methods

### DIFF
--- a/src/lds/continuous_latents.jl
+++ b/src/lds/continuous_latents.jl
@@ -1,6 +1,10 @@
 #=============================================================================
 Continuous (Linear Gaussian) latents
 
+    Log-Likelihood kernels: stateloglikelihood!(cc, dxt, tmp, x, t, lds[, ux])
+                            observationloglikelihood!(cc, b1, b2, x, y, t, lds[, uy])
+                            joint_loglikelihood!(ll, ws, cc, lds, x, y)
+
     E-Step: Q_state!(sws, lds, suf)
 
     M-Step: update_initial_state_mean!(lds, suf)
@@ -8,6 +12,132 @@ Continuous (Linear Gaussian) latents
             update_A_b!(lds, suf, sws)
             update_Q!(lds, suf, sws)
 =============================================================================#
+
+"""
+    LDSLikelihoodCache{T}
+
+Any container of Cholesky-wrapped covariances (`Q_PD` / `P0_PD` / `R_PD`) and
+cached log-likelihood normalizers (`cP0` / `cQ` / `cR`) that the likelihood
+kernels can consume: a per-trial `SmoothWorkspace` (filled by
+`compute_smooth_constants!`) or a per-SLDS-component `LDSConstantCache`
+(filled by `compute_slds_constants!`).
+"""
+const LDSLikelihoodCache{T} = Union{LDSConstantCache{T},SmoothWorkspace{T}}
+
+# Normalizer accessors — `LDSConstantCache` stores plain scalars (mutable
+# struct), `SmoothWorkspace` stores RefValues (immutable struct).
+_cP0(cc::LDSConstantCache) = cc.cP0
+_cQ(cc::LDSConstantCache) = cc.cQ
+_cR(cc::LDSConstantCache) = cc.cR
+_cP0(ws::SmoothWorkspace) = ws.cP0[]
+_cQ(ws::SmoothWorkspace) = ws.cQ[]
+_cR(ws::SmoothWorkspace) = ws.cR[]
+
+"""
+    stateloglikelihood!(cc, dxt, tmp, x, t, lds[, ux])
+
+State-model (prior/transition) contribution to the complete-data log-likelihood
+at timestep `t`:
+
+- `t == 1`: `cP0 - 0.5‖P0^{-1/2}(x_1 - x_0)‖²`
+- `t ≥ 2`:  `cQ - 0.5‖Q^{-1/2}(x_t - A x_{t-1} - b - B u_{t-1})‖²`
+
+This term is identical for every observation model (Gaussian, Poisson, ...);
+only the emission term differs, so `joint_loglikelihood!` implementations share
+this helper instead of duplicating the prior/transition math.
+
+`cc` is an [`LDSLikelihoodCache`](@ref) holding the Cholesky factors and
+normalizers; `dxt` and `tmp` are `latent_dim` scratch vectors (overwritten).
+Pass `ux` (state inputs, `t`-indexed like `x`) to include the `B u_{t-1}`
+term; `nothing` (default) or a zero-row matrix skips it.
+"""
+function stateloglikelihood!(
+    cc::LDSLikelihoodCache{T},
+    dxt::AbstractVector{T},
+    tmp::AbstractVector{T},
+    x::AbstractMatrix{T},
+    t::Int,
+    lds::LinearDynamicalSystem{T0,S,O},
+    ux::Union{Nothing,AbstractMatrix}=nothing,
+) where {T<:Real,T0<:Real,S<:GaussianStateModel{T0},O<:AbstractObservationModel{T0}}
+    A = lds.state_model.A
+    b = lds.state_model.b
+    x0 = lds.state_model.x0
+
+    if t == 1
+        @views dxt .= x[:, 1] .- x0
+        ldiv!(cc.P0_PD[].chol.U', dxt)
+        return _cP0(cc) - T(0.5) * sum(abs2, dxt)
+    else
+        @views mul!(tmp, A, x[:, t - 1])
+        if ux !== nothing
+            @views mul!(tmp, lds.state_model.B, ux[:, t - 1], one(T), one(T))
+        end
+        @views tmp .= x[:, t] .- tmp .- b
+        ldiv!(cc.Q_PD[].chol.U', tmp)
+        return _cQ(cc) - T(0.5) * sum(abs2, tmp)
+    end
+end
+
+"""
+    observationloglikelihood!(cc, buf1, buf2, x, y, t, lds[, uy])
+
+Emission-model contribution `log p(y_t | x_t)` to the complete-data
+log-likelihood at timestep `t`. Dispatches on the observation model type `O`
+(via `lds::LinearDynamicalSystem{T,S,O}`) — a custom observation model plugs
+into `joint_loglikelihood!` by adding a method here, without having to
+reimplement (or duplicate) the shared `stateloglikelihood!` term.
+
+Uniform interface for all observation models:
+- `cc`: an [`LDSLikelihoodCache`](@ref) with Cholesky factors / normalizers
+  (unused by models whose emission term needs no covariance, e.g. Poisson).
+- `buf1`, `buf2`: two `obs_dim` scratch vectors (overwritten); each model uses
+  what it needs (Gaussian: residual; Poisson: linear predictor + rate).
+- `uy` (optional): observation inputs, `t`-indexed like `y`; `nothing` or a
+  zero-row matrix skips the `D u_t` term.
+
+See the `GaussianObservationModel` / `PoissonObservationModel` methods in
+`fit_LDS.jl` / `fit_PLDS.jl` for the pattern to follow.
+"""
+function observationloglikelihood! end
+
+"""
+    joint_loglikelihood!(ll, ws, cc, lds, x, y)
+
+Per-timestep complete-data log-likelihood for a single SLDS component:
+`ll[t] = log p(y_t | x_t) + log p(x_t | x_{t-1})` (or `+ log p(x_1)` at
+`t == 1`). Generic over the observation model — the emission term comes from
+`observationloglikelihood!`, so supporting a new observation model here only
+requires a new `observationloglikelihood!` method, not a new
+`joint_loglikelihood!` method.
+
+Notes:
+- Normalization terms (logdet + log(2π)) are included. These are constant w.r.t.
+  `x`, but **not** constant across SLDS discrete states when `Q`/`R` differ by
+  state.
+"""
+function joint_loglikelihood!(
+    ll::AbstractVector{T},
+    ws::SLDSSmoothWorkspace{T},
+    cc::LDSConstantCache{T},
+    lds::LinearDynamicalSystem{T,S,O},
+    x::AbstractMatrix{T},
+    y::AbstractMatrix{T},
+) where {T<:Real,S<:GaussianStateModel{T},O<:AbstractObservationModel{T}}
+    tsteps = size(y, 2)
+    @assert length(ll) == tsteps
+
+    dxt = ws.dxt
+    tmp = ws.tmp1
+
+    for t in 1:tsteps
+        ll_t = observationloglikelihood!(cc, ws.z, ws.λ, x, y, t, lds)
+        ll_t += stateloglikelihood!(cc, dxt, tmp, x, t, lds)
+        ll[t] = ll_t
+    end
+
+    return ll
+end
 
 """
     Q_state!(ws, lds, E_z, E_zz, E_zz_prev, ux)
--- a/src/lds/continuous_latents.jl
+++ b/src/lds/continuous_latents.jl
@@ -33,6 +33,20 @@ _cP0(ws::SmoothWorkspace) = ws.cP0[]
 _cQ(ws::SmoothWorkspace) = ws.cQ[]
 _cR(ws::SmoothWorkspace) = ws.cR[]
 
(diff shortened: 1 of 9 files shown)


rsenne:
hmm apparently i added some allocations. Gonna see if claudius can fix for me

src/lds/continuous_latents.jl
```diff
@@ -9,6 +13,146 @@ Continuous (Linear Gaussian) latents
...
+    chol::Cholesky{T,Matrix{T}}, v::StridedVector{T}
+) where {T<:LinearAlgebra.BlasFloat}
+    if chol.uplo === 'U'
+        LinearAlgebra.LAPACK.trtrs!('U', 'T', 'N', chol.factors, v)
+    else
+        LinearAlgebra.LAPACK.trtrs!('L', 'N', 'N', chol.factors, v)
+    end
+    return v
+end
+_whiten!(chol::Cholesky, v::AbstractVector) = ldiv!(chol.U', v)
```
harrisonritz, line 48: FYI PDMats has a bunch of whitening functions:

https://github.com/JuliaStats/PDMats.jl/blob/1772c2fb9533e9d98e62060b28b99d55a2c44724/src/pdmat.jl#L141

src/lds/continuous_latents.jl
```diff
@@ -9,6 +13,146 @@ Continuous (Linear Gaussian) latents
...
+term; `nothing` (default) or a zero-row matrix skips it.
+"""
+function stateloglikelihood!(
+    cc::LDSLikelihoodCache{T},
+    dxt::AbstractVector{T},
+    tmp::AbstractVector{T},
+    x::AbstractMatrix{T},
+    t::Int,
+    lds::LinearDynamicalSystem{T0,S,O},
+    ux::Union{Nothing,AbstractMatrix}=nothing,
```
harrisonritz, line 75: Splines PR now defaults ux/uy to zero array. Is that consistent with `nothing` type? Would this be helpful to avoid Union splitting?

harrisonritz: Which also begs the question whether we should have a method that passes `Data` instead of just the parts? Or always require `Data`?

src/lds/continuous_latents.jl
```diff
@@ -9,6 +13,146 @@ Continuous (Linear Gaussian) latents
...
+    tmp = ws.tmp1
+
+    for t in 1:tsteps
+        ll_t = observationloglikelihood!(cc, ws.z, ws.λ, x, y, t, lds)
+        ll_t += stateloglikelihood!(cc, dxt, tmp, x, t, lds)
+        ll[t] = ll_t
+    end
+
+    return ll
+end
```
harrisonritz, line 154: in theory, this should go in `guassian_observations` and `poisson_observations`, but looks like its more general, eh?

maybe we should split off a `likelihood.jl` script that has the joint + marginal likelihood functions?

src/lds/fit_LDS.jl
```diff
@@ -29,158 +29,67 @@ performance, with careful attention to memory allocation and multi-threading.
...
+Built from the shared `stateloglikelihood!` / `observationloglikelihood!`
+kernels; requires `compute_smooth_constants!(ws, lds)` to have been called.
+`ux` / `uy` are optional control inputs (`nothing` or zero-row matrices skip
+the `B u` / `D uy` terms).
 """
 function joint_loglikelihood!(
     ws::SmoothWorkspace{T},
     x::AbstractMatrix{T},
     lds::LinearDynamicalSystem{T0,S,O},
     y::AbstractMatrix{T0},
```
harrisonritz, line 49: Have I just not noticed T0 before -- is this to handle Int poisson observations or something? In Spline PR's `Data` init, I set ux/uy to be the same type as y; doens't uy at least require that?

harrisonritz: should we add T0 to lds{T,S,O}? we can do better than T and T0 -- `Tobs` and `Tlat`?
I'm still confused which type is for which part of the model: ux::T0, but x::t?

harrisonritz: Does this mean that we have to construct `lds` and `Data` together? Or fix latents to Float64?

src/lds/workspaces.jl
```diff
@@ -90,21 +95,21 @@ mutable struct SufficientStatistics{T<:Real}
...
+    dyn_yy::Base.RefValue{DensePDMat{T}}
 
     # observation model
     obs_n::T
-    obs_xx::Base.RefValue{PDMat{T,Matrix{T}}}
+    obs_xx::Base.RefValue{DensePDMat{T}}
     obs_xy::Matrix{T}
-    obs_yy::Base.RefValue{PDMat{T,Matrix{T}}}
+    obs_yy::Base.RefValue{DensePDMat{T}}
 end
```
harrisonritz, line 113: sorry for spirling -- should obs_yy be latent type or observation type?

src/lds/workspaces.jl
```diff
@@ -123,9 +128,14 @@ struct SmoothWorkspace{T<:Real}
     # `compute_smooth_constants!`; downstream code consumes via
     # `ws.R_PD[].chol.U` (triangular factor) and `logdet(ws.R_PD[])`.
     # Mirrors `KalmanWorkspace`'s `Q_PD` / `P0_PD` / `R_PD` pattern.
-    R_PD::Base.RefValue{PDMat{T,Matrix{T}}}      # (obs_dim × obs_dim)
-    Q_PD::Base.RefValue{PDMat{T,Matrix{T}}}      # (latent_dim × latent_dim)
-    P0_PD::Base.RefValue{PDMat{T,Matrix{T}}}     # (latent_dim × latent_dim)
+    R_PD::Base.RefValue{DensePDMat{T}}      # (obs_dim × obs_dim)
+    Q_PD::Base.RefValue{DensePDMat{T}}      # (latent_dim × latent_dim)
```
harrisonritz, line 132: should R be latent type or observation type?

src/lds/workspaces.jl
```diff
@@ -637,6 +663,12 @@ function compute_smooth_constants!(
     ldiv!(P0_chol, ws.x_t)
     ws.x_t .*= -one(T)
 
+    # Log-likelihood normalizers. No R term for Poisson observations.
+    latent_dim = lds.latent_dim
+    ws.cP0[] = -WT(0.5) * (WT(latent_dim) * log(WT(2π)) + logdet(ws.P0_PD[]))
+    ws.cQ[] = -WT(0.5) * (WT(latent_dim) * log(WT(2π)) + logdet(ws.Q_PD[]))
+    ws.cR[] = zero(WT)
```
harrisonritz, line 670: again, do these all share the same types?

src/lds/workspaces.jl
```diff
@@ -955,9 +987,9 @@ struct KalmanWorkspace{T<:Real}
     p_smooth_tt1_shared::Array{T,3}  # (D, D, T)
 
     # Derived constants (refreshed each E-step; M-step mutates Q, R, P0)
-    Q_PD::Base.RefValue{PDMat{T,Matrix{T}}}
-    P0_PD::Base.RefValue{PDMat{T,Matrix{T}}}
-    R_PD::Base.RefValue{PDMat{T,Matrix{T}}}
+    Q_PD::Base.RefValue{DensePDMat{T}}
+    P0_PD::Base.RefValue{DensePDMat{T}}
+    R_PD::Base.RefValue{DensePDMat{T}}
```
harrisonritz, line 992: all same types?

test/LinearDynamicalSystems/PoissonLDS.jl
```diff
@@ -685,3 +685,47 @@ function test_poisson_low_rate_recovery()
...
+        obs_model=om,
+        latent_dim=D_lat,
+        obs_dim=p_obs,
+        fit_bool=fill(true, 6),
+    )
+
+    x = randn(rng, D_lat, T_steps)
+    λ = exp.(C_nd * x .+ d_nd)
+    y = Float64.(rand.(Ref(rng), Poisson.(λ)))
+
```
harrisonritz, line 718: oh, so y is not a different type than the latents... so why have T and T0?

src/lds/fit_LDS.jl
```diff
@@ -29,158 +29,67 @@ performance, with careful attention to memory allocation and multi-threading.
...
-        ldiv!(temp_solve_R, R_U, temp_dy)
-        ll_t += cR - T(0.5) * sum(abs2, temp_solve_R)
-
+        ll_t = observationloglikelihood!(ws, ws.temp_dy, ws.temp_solve_R, x, y, t, lds, uy)
+        ll_t += stateloglikelihood!(ws, ws.temp_dx, ws.temp_solve_Q, x, t, lds, ux)
         ll_vec[t] = ll_t
     end
 
     return ll_vec
 end
```
harrisonritz, line 64: guessing at the right types here

harrisonritz: but see other spin-outs about type matching.

src/lds/fit_LDS.jl
```diff
@@ -29,158 +29,67 @@ performance, with careful attention to memory allocation and multi-threading.
...
-    temp_dy = ws.temp_dy
-    temp_solve_Q = ws.temp_solve_Q
-    temp_solve_R = ws.temp_solve_R
-
-    latent_dim = lds.latent_dim
-    obs_dim = lds.obs_dim
-
-    cP0 = -T(0.5) * (T(latent_dim) * log(T(2π)) + logdet(ws.P0_PD[]))
-    cQ = -T(0.5) * (T(latent_dim) * log(T(2π)) + logdet(ws.Q_PD[]))
-    cR = -T(0.5) * (T(obs_dim) * log(T(2π)) + logdet(ws.R_PD[]))
```
harrisonritz, line 54: should this be acopy? is ws.ll_vec used for anything else?

rsenne:
I realized this PR is targeting main not dev. I'll fix

harrisonritz (commented):
I love the idea here, and the oppertunity to consolidate these likelihood functions. see comment for whether these are general enough to go into a likelihood.jl script (I think they might be)

However I'm confused about a few things:


1. sorry for spinning out about all these types, but I don't understand what T0/T/U/WT refer to, and what type is stored in SmoothWorkspace/lds. If we need multiple types, then I think they should have clear + consistent naming (eg T_latent, T_obs, T_smooth (?), etc). But shouldnt the latents and observations be the same type (both floats)?

2. what is the use case for joint_loglikelihood? IIUC its not called in `elbo!`, and isn't part of the fitting process. If this is supposed to be called after the model is already fit, then I think it needs to take the data in the same format that its passed to fit! -- eg a vector of multiple trials. 

Are we avoiding passing a `Data` struct to `fit!` to match StatsAPI? Maybe it would be good to unify the user-facing functions (fit!, loglikelihood, joint_loglikelihood, simulate, etc) with a standard data format -- I'd say that `Data` would make the most sense, but happy for something else.

rsenne:
Gonna give one quick response here then I'll give a more thorough one later

> sorry for spinning out about all these types, but I don't understand what T0/T/U/WT refer to, and what type is stored in SmoothWorkspace/lds. If we need multiple types, then I think they should have clear + consistent naming (eg T_latent, T_obs, T_smooth (?), etc). But shouldnt the latents and observations be the same type (both floats)?

The real answer is historicity. A lot of this stems from me not understanding typing when I first built everything and not understanding how the Julia compiler really worked. In reality--we should be able to make this typing way more generic. To do this we need to create overloads of `Base.eltype()`.

Realistically this should be done for the new release as its a major improvement on the dev side of things. For example, the following function

```julia
function joint_loglikelihood(
    x::AbstractMatrix{U}, plds::LinearDynamicalSystem{T,S,O}, y::AbstractMatrix{T}) where {U<:Real,T<:Real,S<:GaussianStateModel{T},O<:PoissonObservationModel{T}}
```

really could be:


```julia
function joint_loglikelihood(
    x::AbstractMatrix, plds::LinearDynamicalSystem{S,O}, y::AbstractMatrix) where {S<:GaussianStateModel,O<:PoissonObservationModel}
```

I think the typing with the Observation/state model still makes sense for dispatch reasons however. But there is many cases across the board where realistically typing isn't handled well

rsenne:
New issue opened: #137

src/lds/continuous_latents.jl
rsenne: I'll take a look thanks!

src/lds/continuous_latents.jl
rsenne: A zero array would not be compatible with a `nothing` type

src/lds/continuous_latents.jl
rsenne: Yes this is the backbone of the loglikelihoods that should in theory be consistent with any model type where we assume the latent process is gaussian. I figured `continuous_latents.jl` would be a good home for it as that is what i guess is common to both but unopposed to a new file!

src/lds/continuous_latents.jl
rsenne: i'll respond to other points in main thread!

src/lds/fit_LDS.jl
rsenne: Not super pertinent but a lot of this was related to AD compat what i mention in the main thread is a waaaay better solution. I knew this needed to be taken care of eventually

src/lds/fit_LDS.jl
rsenne: not sure i follow?

rsenne:
> what is the use case for joint_loglikelihood? IIUC its not called in elbo!, and isn't part of the fitting process. If this is supposed to be called after the model is already fit, then I think it needs to take the data in the same format that its passed to fit! -- eg a vector of multiple trials.

1. Its partially a historical artifact
2. It is the optimization target of the Paninski method. The reason LDS isn't using it is because we are guaranteed to converge in a single step so we don't need to use Newton's method (technically we still are doing Newton's method i suppose). When i did all the optimizations i removed it for speed

4. For the PLDS its a bit more of a complicated thing in that it should be whats optimized on but apparently i removed the normalizer as it was too slow (going off of what my comment is). I actually find that somewhat hard to believe so I might try seeing if actually that is even accurate.

So tl;dr for a generic model here the `joint_loglikelihood` would be the optimization target for smoothing. Another useful, but tangential way it can be used is for testing. We can in theory use monte carlo samples to estimate the elbo as its an unbiased estimator so easy way to check that our elbos are correct

> Are we avoiding passing a Data struct to fit! to match StatsAPI? Maybe it would be good to unify the user-facing functions (fit!, loglikelihood, joint_loglikelihood, simulate, etc) with a standard data format -- I'd say that Data would make the most sense, but happy for something else.

Unifying the interface would be ideal. Sorry to turn this question back on you--but, this hadn't really clicked for me until now, why the `Data` struct? Couldn't we just pass matrices/vector of matrices etc? Sorry if we already discussed this

src/lds/continuous_latents.jl
rsenne: I tested this and the PDMats one allocates unfortunately and the performance hit is measurable. Its not huge--but what do you think?

harrisonritz:
ok -- so helpful to have joint_loglikelihood somewhat user-facing as a likelihood surrogate.

For `Data`, I think some reasons for keeping stuff in a struct by default:
- data validation at construction (eg y array -> vector(matrix))
- better memory management (iiuc immutability means it can be stack-allocated)
- more readable functions
- easier type stability
- y/ux/uy are almost always used together

rsenne:
I didn't have time to address this but i opened #139 as a place to talk about more design pieces like this. I'm thinking maybe before we add more features maybe we get all of these API designs cemented? What do you think? My thought process is our discussion has spun out a few issue that feel like more foundational to the package and worth addressing first

Overarching issues
---
- Fix typing
- Unify API
- Decide how data is passed in (e.g., data or?)
- Deduplicate code

Sub issues
---
- Related to some of these having proper gettrs/settrs
- Maybe add workspaces to the models? In theory struct fields are private without gettrs so this in in relation to above.

Just some broad thoughts---looking forward to discussing more!

harrisonritz:
I think that's the right call :) I was flagging stuff at higher level than this PR, was useful for drudging stuff up.

harrisonritz (approved):

PR #136, opened 2026-07-08
Dedupe LDS gradients into shared kernels

rsenne:
Stacked on #135 — review only the top commit; the rest is the base branch. Retarget to `main` after #135 merges.

Deduplicates the per-observation-model `Gradient!` implementations (Gaussian / Poisson / SLDS-weighted) into shared kernels in `continuous_latents.jl`, mirroring the likelihood split from #135:

- `observationgradient!(out, cc, buf, x, y, t, lds[, uy])` — per-model emission gradient dispatch point; a new observation model plugs in with one method.
- Shared `Gradient!` assembles emission + prior/transition factors from the cached Cholesky-derived templates.
- `_transition_residual!` defines the affine-dynamics residual `x_t - A x_{t-1} - b - B u_{t-1}` in exactly one place, shared by the likelihood and gradient kernels.

### DIFF
--- a/src/lds/fit_LDS.jl
+++ b/src/lds/fit_LDS.jl
@@ -24,8 +24,7 @@ Linear Dynamical System (LDS) implementation with Gaussian state and observation
 This module defines functions specific to the Gaussian observation model: log-likelihoods,
 gradients, Hessians, smoothing, the ELBO/sufficient-statistics machinery, and the EM
 fit driver. Model-agnostic helpers (parameter extraction, `initialize_FilterSmooth`)
-live in `common.jl`; sampling lives in `simulate.jl`. The code is optimized for
-performance, with careful attention to memory allocation and multi-threading.
+live in `common.jl`; sampling lives in `simulate.jl`.
 """
 
 """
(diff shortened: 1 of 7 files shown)


src/lds/fit_LDS.jl
```diff
@@ -108,111 +105,27 @@ function observationloglikelihood!(
 end
 
 """
-    Gradient!(ws, lds, y, x)
+    observationgradient!(out, cc, buf, x, y, t, lds[, uy])
```
harrisonritz, line 108: Maybe better-suited for gaussian_observations.jl?

Also, how about `observation_gradient!` and `observation_loglikelihood!` etc, just for readability?

harrisonritz (approved):
Ya looks good. 

Minor suggestions for naming policy "xxx_gradient" and "xxx_loglikelihood". 

Consider putting observation_gradient in "xxx_observations.jl" rather than "fit_xxx.jl"

rsenne:
I'm on board with that. I'll delegate all those changes in the last of this 3 part series

Issue #140, opened 2026-07-09
v0.5.0 umbrella: dedupe, typing, and public API unification


rsenne:
Tracking issue for the internals/API cleanup that should land together for **v0.5.0**. The individual pieces each have their own issue, but they interlock — the dedupe work shrinks the surface the typing and API changes have to touch, and the API unification is the breaking change that sets the version bump — so ordering matters more than it looks from any one issue.

### 1. Code dedupe — #131
Collapse the per-observation-model likelihood/gradient/hessian implementations into shared kernels with one dispatch point per observation model.
- [x] Likelihoods — #135
- [x] Gradients — #136 (stacked on #135, retarget to `dev` after merge)
- [x] Hessians — #141 (stacked on #136, ibid.)

**Why first:** every later step (typing, API) touches these functions. Doing dedupe first means we loosen/rework each computation in one place instead of three.

### 2. Public API unification — #139
One `joint_loglikelihood(lds, x, y; latent_inputs, obs_inputs)` shape: model-first, scalar return, `fit!`-format trials, kwargs matching `fit!`. Same treatment should extend to the other user-facing functions where argument order / return shape currently varies.
- [x] Settle the open design question: plain kwargs vs. the `Data`-struct direction from the #135 review thread (blocks #127, which modifies `Data`)
- [ ] Implement over the deduped kernels
- [ ] CHANGELOG migration notes (this is the breaking change → v0.5.0)

### 3. Typing refactor — #137
`Base.eltype` overloads + type promotion instead of over-restrictive signatures; better interop with non-standard matrix types.
- [ ] Audit signatures after dedupe (far fewer to touch then)
- Can slip past 0.5.0 if it drags — it's additive-loosening, not breaking

### 4. Threading — #138
Evaluate OhMyThreads.jl vs. current `Threads.@spawn` usage.
- [x] Benchmark before committing; no API impact, safe to defer to 0.5.x (see #143)

## Downstream / related (not part of this track, but sequenced by it)
- #127 (input bases) — blocked on the #139 `Data` decision; changes requested
- #126 (null models) — wants #127's inputs machinery; changes requested
- #129 (SSID init) — additive, independent; CI red + conflicts, fine as 0.5.x

## Rough order

#135 → #136 → hessian dedupe → #139 decision + implementation → #127 → #126, then CHANGELOG + version bump to 0.5.0 and release from `dev`. #137/#138/#129 ride along if ready, otherwise 0.5.x.

[...]

PR #141, opened 2026-07-09
Start Hessian depduplication.

rsenne:
Just getting this going to keep track

### DIFF
--- a/src/lds/continuous_latents.jl
+++ b/src/lds/continuous_latents.jl
@@ -406,9 +406,11 @@ function Q_state!(
     fill!(sum_mu_t, zero(T))
     fill!(sum_mu_tm1, zero(T))
 
-    # Input-specific accumulators (only allocated when ux_dim > 0). Allocating
-    # 0-element arrays here would still cost an `Array` struct each call,
-    # which adds up to thousands of trivial allocations across a fit.
+    #=
+    Input-specific accumulators (only allocated when ux_dim > 0). Allocating
+    0-element arrays here would still cost an `Array` struct each call,
+    which adds up to thousands of trivial allocations across a fit.
+    =#
     has_input = ux_dim > 0
     sum_u = has_input ? zeros(T, ux_dim) : Vector{T}()
     sum_mu_t_u = has_input ? zeros(T, D, ux_dim) : Matrix{T}(undef, 0, 0)
@@ -447,9 +449,11 @@ function Q_state!(
     mul!(temp, b, ws.tmp1', one(T), one(T))
     mul!(temp, b, b', T(tstep - 1), one(T))
 
-    # Input cross terms (`Bu_{t-1} := B u_{t-1}`). All terms here are
-    # contributions to `Σ_t E[(x_t - A x_{t-1} - b - B u_{t-1})(...)']` that
-    # involve at least one `B u_{t-1}` factor.
+    #=
+    Input cross terms (`Bu_{t-1} := B u_{t-1}`). All terms here are
+    contributions to `Σ_t E[(x_t - A x_{t-1} - b - B u_{t-1})(...)']` that
+    involve at least one `B u_{t-1}` factor.
+    =#
     if ux_dim > 0
         # -= sum_mu_t_u · B'  and  -= B · sum_mu_t_u'
         mul!(temp, sum_mu_t_u, B', -one(T), one(T))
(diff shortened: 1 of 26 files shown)


rsenne:
Just FYI i started to do some comment reformatting across the repo so the diff is a bit inflated--i figure we can knock out a few chores here and there as we do this

rsenne:
For my next trick I'll start to decrease verbosity

src/lds/continuous_latents.jl
```diff
@@ -224,7 +227,7 @@ function Gradient!(
 
     # Middle steps: emission + incoming factor at t + outgoing factor at t + 1
     @views for t in 2:(tsteps - 1)
-        observationgradient!(tmp1, ws, obs_buf, x, y, t, lds, uy)
+        observation_gradient!(tmp1, ws, obs_buf, x, y, t, lds, uy)
```
harrisonritz, line 230: God saw all the snake_case that he had made, and it was very good

src/lds/continuous_latents.jl
```diff
@@ -253,6 +256,98 @@ function Gradient!(
...
+- `buf1`, `buf2`: two `obs_dim` scratch vectors (overwritten); each model uses
+  what it needs (Gaussian: none; Poisson: linear predictor + rate).
+- `y` is unused by the built-in models (Gaussian/Poisson-canonical curvature is
+  observation-independent) but part of the interface for models whose curvature
+  depends on `y`.
+
+See the `GaussianObservationModel` / `PoissonObservationModel` methods in
+`fit_LDS.jl` / `fit_PLDS.jl` for the pattern to follow.
+"""
+function observation_hessian! end
```
harrisonritz, line 286: Just curious, why define empty method here? We'll never call it -- for continuity when reading the script?

src/lds/continuous_latents.jl
```diff
@@ -253,6 +256,98 @@ function Gradient!(
...
+- `H_sub[i] = Q⁻¹A`, `H_super[i] = (Q⁻¹A)'` for all i
+- `H_diag[1] = -A'Q⁻¹A - P0⁻¹`
+- `H_diag[t] = -A'Q⁻¹A - Q⁻¹` (middle), `H_diag[T] = -Q⁻¹`
+
+Uses the templates cached on `cc` by `compute_smooth_constants!` /
+`compute_slds_constants!`. Overwrites the diagonal blocks — callers add the
+emission curvature afterwards via `observation_hessian!`. Requires `tsteps ≥ 2`
+(matching the Newton smoother's contract).
+"""
+function _state_hessian_blocks!(btd, cc::LDSLikelihoodCache{T}, tsteps::Int) where {T<:Real}
```
harrisonritz, line 303: It's very satisfying how tidy these are now

src/lds/fit_LDS.jl
```diff
@@ -323,20 +250,24 @@ function smooth!(
...
+    Equal-length fast path: the BT Hessian (and its inverse) is observation-
+    independent, so the smoothed covariance is identical across trials. Run
+    the cov pass once on `sws_pool[1]`, alias each FilterSmooth's
+    `p_smooth` / `p_smooth_tt1` to the shared storage, then do gradient-and-
+    solve per trial in parallel.
+    =#
     T1 = size(y[1], 2)
     all_equal = all(yt -> size(yt, 2) == T1, y)
 
     if all_equal
```
harrisonritz, line 263: flag for https://github.com/depasquale-lab/StateSpaceDynamics.jl/issues/140 : 

wonder if this forking would benefit from a method for`Data`::FixedLength, especially if `y::Array` is more efficient in general than `y::Vector(Matrix)`. 

Also depends on how much duplication that would require.

Don't have a great intuition for where function barriers are helpful in general, potentially good here?

src/lds/gaussian_observations.jl
```diff
@@ -215,3 +225,77 @@ function update_R!(
...
+    cc::LDSLikelihoodCache{T},
+    ::AbstractVector{T},
+    ::AbstractVector{T},
+    x::AbstractMatrix{T},
+    y::AbstractMatrix{T0},
+    t::Int,
+    lds::LinearDynamicalSystem{T0,S,O},
+    α::T=one(T),
+) where {T<:Real,T0<:Real,S<:GaussianStateModel{T0},O<:GaussianObservationModel{T0}}
+    @. out += α * cc.yt_given_xt
```
harrisonritz, line 299: Feels like there must be a BLAS function we could alias for this? Maybe this is as fast as it goes?

harrisonritz (approved):
Generally looks good!

src/lds/continuous_latents.jl
rsenne: This is the standard Julia idiom for declaring a generic-function interface point: the empty stub gives `observation_hessian!` an existence and a docstring (so ?observation_hessian! works) before any concrete methods exist

An added benefit too is if someone adds a new observation model and forgets to implement this method, calling it throws a clear `MethodError: no method matching observation_hessian!(...)`

I think this will be more relevant alter when we make it easier to add generic methods

src/lds/gaussian_observations.jl
rsenne: I think so: `axpy!(α, cc.yt_given_xt, out)` probably worth benchmarking at some point

PR #144, opened 2026-07-11
Rework Workspaces

rsenne:
Beginning of restructuring workspaces so that we can unify many dev facing functions. This is in service of #140.

So far what this does is it makes a `SmoothedWorkspace` more modular (~90 fields was a mess lets be real.) into now:

- `BlockTridiagonalWorkspace`
- `SmoothConstants`
- `NewtonBuffers`
- `RegressionBuffers`
- `ElboBuffers`
- `TDAggBuffers`
- `BatchedBuffers`

Thing I'd like to further do:

- [x] ~Add gettrs for readability (e.g., sw.soemthing.somethingelse is hard to read)~ (After perusing this is unidiomatic)
- [x] Unify the function handles
- [ ] profit ?


rsenne:
@harrisonritz prelim thoughts on this?

harrisonritz:
Ya good idea -- IIUC there's ~0 penalty for struct depth, compiler flattens everything. Looks like some good tidy-up as well, thank you!

rsenne:
LMK if you think there is any other workspace work that should go here! i'm sure more will happen but if there is any gaping holes

src/lds/fit_LDS.jl
```diff
@@ -159,8 +153,8 @@ function smooth!(
     compute_smooth_constants!(sws, lds)
 
     # Workspace buffers may be sized for a longer trial; take active-length views.
-    X0 = view(sws.X₀, 1:n_active)
-    grad_vec = view(sws.grad_vec, 1:n_active)
+    X0 = view(sws.opt.X₀, 1:n_active)
```
harrisonritz: Hm have we always been using `X_0` as a param? 

TBH, I kind of hate the Julia asci parameters, because you don't have a keyboard key for them (eg now do i type X_0?). Looks nice, but worse usability for users/devs IMO. Just my opinion.

src/lds/fit_SLDS.jl
```diff
@@ -408,8 +409,10 @@ function Hessian_blocks!(
         fill!(H_super[t], zero(T))
     end
 
-    z = ws.z
-    λ = ws.λ
+    # Two obs_dim scratch vectors for observation_hessian! (Poisson writes the
+    # linear predictor and rate into them; Gaussian ignores both).
+    z = ws.opt.dyt
+    λ = ws.opt.temp_dy
```
harrisonritz, line 415: Confirm these two assignment are the same as before (different variables names)

src/lds/poisson_observations.jl
```diff
@@ -187,13 +187,13 @@ arguments are unused (no covariance term; Poisson observation inputs are not
 supported).
 """
 function observation_loglikelihood!(
-    ::LDSLikelihoodCache{T},
+    ::SmoothConstants{T},
     z::AbstractVector{T},
     λ::AbstractVector{T},
```
harrisonritz, line 192: If z/lambda these aren't workplaces, maybe should come after x/y?

src/lds/workspaces.jl
```diff
@@ -310,235 +504,41 @@ function SmoothWorkspace(
...
-        batched_dxt_next,
-        batched_dyt,
-        batched_tmp1,
-        batched_tmp2,
-        batched_tmp3,
-        batched_y,
-        batched_ux,
-        batched_uy,
-        batched_data_valid,
+        BlockTridiagonalWorkspace(T, latent_dim, tsteps),
```
harrisonritz: Maybe have a one-liner comment for each of these sub-structs, so someone climbing the structure tree can get a high-level idea for what they do. Documentation is very good above

src/lds/workspaces.jl
```diff
@@ -562,328 +562,170 @@ function compute_smooth_constants!(
...
+    `WT === T` and avoids a Julia 1.10 `Cholesky` convert-method gap
+    when `WT !== T` (ForwardDiff path).
     =#
     Q_w = convert(Matrix{WT}, Q)
     P0_w = convert(Matrix{WT}, P0)
-    ws.Q_PD[] = PDMat(Symmetric(Q_w))
-    ws.P0_PD[] = PDMat(Symmetric(P0_w))
-    Q_chol = ws.Q_PD[].chol
-    P0_chol = ws.P0_PD[].chol
+    cc.Q_PD = PDMat(Symmetric(Q_w))
```
harrisonritz: Force symmetric with `Symmetricize!`?

harrisonritz (approved):
Looks great, very minor comments

src/lds/fit_LDS.jl
rsenne: X0 is fine with me as a name. FWIW one can type asci characters in vscode using "\" e.g., X\_0 followed by a tab -> X₀. I do like to use them when it makes sense, this example i agree is unnecessary

src/lds/fit_SLDS.jl
rsenne: yes these are equivalent / these are intentional I realized i could reuse these for poisson

src/lds/poisson_observations.jl
rsenne: these are mutated to style guide says they go first / also they are scratch buffers so i guess they are workspace adjacent

rsenne:
unclear why we do better on memory for gaussian now but i am not complaining lol

PR #145, opened 2026-07-14
Fix/slds elbo

rsenne:
Several fixes to the SLDS model. I realized there were a few bugs in the implementation:

1.) The ELBG was wrong (had an incorrect sign etc.) Now test a K=1 SLDS == LDS ELBO
2.) I sampled the marginals--not the true q(x) so really i was doing a mean field sampling

This hence improves the SLDS Implementation and adds more correctness testing

### DIFF
--- a/docs/examples/SLDS.jl
+++ b/docs/examples/SLDS.jl
@@ -43,7 +43,7 @@ obs_dim = 10
 K = 2
 
 A_hmm = [
-    0.92 0.08;
+    0.92 0.08
     0.06 0.94
 ]
 πₖ = [1.0, 0.0]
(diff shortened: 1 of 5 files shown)


rsenne:
Literally zero rush to review this--just getting it going

docs/examples/SLDS.jl
```diff
@@ -140,32 +153,42 @@ p_elbo = plot(elbos; xlabel="iteration", ylabel="ELBO",
 #     This section reaches into non-exported internals (`initialize_FilterSmooth`,
 #     `SLDSDiscreteLayer`, `estep!`, and friends) because the package does not
 #     yet expose a public posterior-decoding API for SLDS. These names may
-#     change between releases. The `obs_inputs`/`latent_inputs` values below
+#     change between releases. The `obs_seq`/`control_seq` values below
```
harrisonritz, line 156: Isn't *_inputs used elsewhere?

harrisonritz: I see, answered below (it's just for HMM). Weird.

harrisonritz: Back to being confused. If obs_seq are an internal thing for HMM.jl, then do they need to be exposed in `fit!`? Maybe if they are time-steps (?) and not inputs, then maybe a name that more clearly differentiates them from obs_inputs (though soon to be uy). Maybe they are the prophecized `uz`?

harrisonritz (approved):
A little over my head, but the gist makes sense.

Love the K=1 SLDS == LDS check, very important. Code coverage looks like we need another test or two to catch the new additions (or route something existing through these functions).

harrisonritz:
Btw -- what is happening in this test? Should this be passing? do we just need a bigger dataset?image

rsenne:
> Btw -- what is happening in this test? 

Yeah this used to work fine--but StableRNGs are only guaranteed stable on the same Julia version. So at some point we just started being unlucky. I'll bump up the data

docs/examples/SLDS.jl
rsenne: yes---we still need to pass it down to HMM.jl so that we handle the trialized cases

rsenne: wait i'm conflating two things i think let me look quick

rsenne: Oh yeah I remember now. This is entirely a hack i did to get this to work and I lowkey hate it. In a future version i will make this a proper HMM type as opposed to just getting this to "work". All this does is indexes into a precomputed loglikelihood. Im basically abusing the HMM controil interface

PR #147, opened 2026-07-14
Fix 4 Random Issues

rsenne:
Should be 4 separate PRs but well.

Fixes:
- Incorrect M-step in PPCA (need to use W_new for sigma^2 calc)
- Line-search cubic interpolation only ever use minimizer. Realistically this doesn't matter as we only minimize there but idk technically maximization is an option so I'd rather it just be correct
- `validate_probvec` could give wrong results for lower precision types e.g., `Float32`s
- tol_PD method error is fixed

### DIFF
--- a/src/utils/validation.jl
+++ b/src/utils/validation.jl
@@ -393,26 +393,13 @@ function validate_SLDS(slds::SLDS)
         throw(DimensionMismatchError("size(A, 1) vs number of LDSs", lds_count, k))
     end
 
-    # Validate transition matrix rows
+    # Validate transition matrix rows and the initial distribution. Delegating to
+    # validate_probvec keeps a single source of truth for the type-scaled
+    # tolerance (see A12) instead of re-hardcoding 1.0/atol=1e-10 here.
     for i in 1:k
-        row = slds.A[i, :]
-        sum_val = sum(row)
-        has_neg = any(x -> x < 0, row)
-        has_gt1 = any(x -> x > 1, row)
-
-        if !isapprox(sum_val, 1.0; atol=1e-10) || has_neg || has_gt1
-            throw(InvalidProbabilityVectorError("A[$i, :]", sum_val, has_neg, has_gt1))
-        end
+        validate_probvec(@view(slds.A[i, :]); name="A[$i, :]")
     end
-
-    # Validate initial distribution
-    sum_val = sum(slds.πₖ)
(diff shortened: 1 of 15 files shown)


harrisonritz:
I think you need a test where the SLDS has a prior

src/utils/validation.jl
```diff
@@ -393,26 +393,13 @@ function validate_SLDS(slds::SLDS)
...
-    # Validate initial distribution
-    sum_val = sum(slds.πₖ)
-    has_neg = any(x -> x < 0, slds.πₖ)
-    has_gt1 = any(x -> x > 1, slds.πₖ)
-
-    if !isapprox(sum_val, 1.0; atol=1e-10) || has_neg || has_gt1
-        throw(InvalidProbabilityVectorError("πₖ", sum_val, has_neg, has_gt1))
+        validate_probvec(@view(slds.A[i, :]); name="A[$i, :]")
     end
+    validate_probvec(slds.πₖ; name="πₖ")
```
harrisonritz, line 402: Another ASCII that I noticed -- how do you even type _k? Can just call pi? 

Disregard if better to be consistent with hmm.jl

test/Utilities/Utilities.jl
```diff
@@ -224,6 +224,27 @@ function test_valid_Σ()
     @test !valid_Σ([2.0 0.3; 0.7 1.0])
 end
 
+function test_tol_PD_type_preservation()
```
harrisonritz, line 227: Good catch

harrisonritz (approved):
Looks good. 

Minor, let's try to branch off dev/main so the PRs don't compound.. maybe you were trying to avoid a tricky merge

rsenne:
> Minor, let's try to branch off dev/main so the PRs don't compound.. maybe you were trying to avoid a tricky merge

Yes! This one is off dev--the other two are interrelated so one of them is a stacked PR, I've found thats helpful for merges where one depends on the other similar to the hessian/gradient/likelihood ones

Oh wait i see your point. this was intended to be off dev. Leaving this comment up so i feel shame

src/utils/validation.jl
rsenne: You can type out subscripts in vscode similar to to the other nonstandard like mu or pi! So one can do for this example: \pi (hit enter or tab) then you can do π\_k and hit enter again and you'll get the subscript

PR #150, opened 2026-07-15
Fix/standardize inputs

rsenne:
Shoutout to vscode find + replace on this one

Also starts the changelog. Once this merges I will start a changelog branch that we can update as PRs are merged to keep track


harrisonritz (approved):
Did this really reduce allocations? This is feeding my short-variable-names-save-memory paranoia

rsenne:
> reduce allocations

Sadly i think this is just because its currently comparing to main (though I'm not sure why)

PR #152, opened 2026-07-15
Fix/delete kf

rsenne:
Deletes Joseph Smoother and EM for the dual implementation of the KF. The information form filter / predictions moved into the BTD pathway

Also--all `loglikelihoods` -> `StatsAPI.loglikelihood`

Once this is in I'll move onto actually unifying things


rsenne:
The `PD_tol` was messing up the MLs in the docs example i guess because the perturbations of different rotations can be more/less extreme. To make the ML calc more faithful I removed, but maybe there is another use case for it? Though it seems `Symmetrize!` + PDMats etc. is working okay?

rsenne:
@harrisonritz all tests/things look good. lmk if i deleted anything i shouldn't have

harrisonritz (approved):
Looks good

Issue #139, opened 2026-07-09
Unify the joint_loglikelihood API (model-first, scalar, fit!-format trials)


rsenne:
Spun out of the #135 review discussion with @harrisonritz about unifying the user-facing functions and the `fit!` data format.

## Current `joint_loglikelihood` surface

| Where | Signature | Returns |
|---|---|---|
| Gaussian | `joint_loglikelihood!(ws, x, lds, y[, ux, uy])` | per-timestep `ws.ll_vec` |
| Gaussian | `joint_loglikelihood(x, lds, y[, ux, uy])` | per-timestep vector |
| Poisson | `joint_loglikelihood!(ws, x, plds, y[, lognorm_t])` | per-timestep view |
| Poisson | `joint_loglikelihood(x, plds, y)` | per-timestep vector |
| Poisson | `joint_loglikelihood(xs::Vector, plds, ys::Vector)` | **scalar** total |
| SLDS (internal) | `joint_loglikelihood!(ll, ws, cc, lds, x, y)` | fills caller's `ll` |

Inconsistencies:
- return type flips between per-timestep vector and scalar depending on input shape
- argument order is `(x, lds, y)` while `fit!` / `loglikelihood` / `smooth` are model-first
- Gaussian has no multi-trial method; Poisson takes no inputs
- inputs are positional `ux`/`uy` here but `latent_inputs`/`obs_inputs` kwargs on `fit!`

Notably, every caller outside SLDS internals (tests, docs) immediately does `sum(...)` — nothing public consumes the per-timestep vector.

## Proposal

Public — one function, one shape, scalar:

```julia
joint_loglikelihood(lds, x, y; latent_inputs=nothing, obs_inputs=nothing) -> Real
```

- model-first, matching `fit!` / `loglikelihood` / `smooth`
- methods for a single trial (matrices) and for `fit!`-format trials (vectors of matrices)
- kwargs named identically to `fit!`

Internal — one per-timestep worker per observation model (this is already the shape on both the Gaussian and Poisson sides after #135):

```julia
joint_loglikelihood!(ws, x, lds, y, ...)  # per-timestep into ws.ll_vec
```

The public function is `sum` over a worker with a fresh workspace; the PLDS Newton line-search objective is `sum` over a worker with a pooled workspace and hoisted normalizer. The SLDS internal variant can fold into the same shape in a later pass.

## Notes

[...]

Issue #137, opened 2026-07-08
Refactor Typing


rsenne:
This package should have overloads of `Base.eltype` and should use type promotion. Right now we are too restrictive on types and it's a headache from the development perspective but also reduces interoperability with other matrix types etc.

PR #153, opened 2026-07-16
Begin API unification

rsenne:
- Implements the private `Data` `struct` as discussed in #139
- Begins to unify the backend / user facing interfaces
- Adds a more informative error for SLDS
- Adds public allocating elbo
- Adds validators of `Data`

This does not fully resolve our API issues, but i figure its best to do this in small bursts.


rsenne:
huzzah its fixed

rsenne:
@harrisonritz will you be able to take a peek at this soon? (note the diff is a bit big as i merged #154 into this)

src/lds/fit_LDS.jl
```diff
@@ -599,20 +641,19 @@ function _smooth_mean_only_batched!(
...
 function smooth!(
     lds::LinearDynamicalSystem{T,S,O},
     tfs::TrialFilterSmooth{T},
     y::AbstractVector{<:AbstractMatrix{T}},
     sws_pool::Vector{SmoothWorkspace{T}},
 ) where {T<:Real,S<:GaussianStateModel{T},O<:GaussianObservationModel{T}}
-    ux = [zeros(T, 0, size(yt, 2)) for yt in y]
-    uy = [zeros(T, 0, size(yt, 2)) for yt in y]
-    return smooth!(lds, tfs, y, sws_pool, ux, uy)
+    return smooth!(lds, tfs, Data(lds, y), sws_pool)
```
harrisonritz: I've found that defining stuff in function calls gives me memory leaks ¯\_(ツ)_/¯
Not sure whether this is the case for structs.

ie I would usually allocate before the call: 

```Julia
data= Data(lds,y)
smooth!(lds, tfs, data, sws_pool)
```

Maybe I was wrong about this! But I had tons of problems with memory leaks in SSA when I ran big jobs on the cluster.

src/lds/fit_SLDS.jl
```diff
@@ -853,6 +853,102 @@ function elbo!(
...
+  matrices (ragged lengths allowed).
+
+Returns a scalar.
+"""
+function elbo(
+    slds::SLDS{T,S,O},
+    y::Union{AbstractMatrix{T},AbstractArray{T,3},AbstractVector{<:AbstractMatrix{T}}};
+    rng::AbstractRNG=Random.default_rng(),
+) where {T<:Real,S<:AbstractStateModel,O<:AbstractObservationModel}
+    # SLDS takes no ux/uy inputs; Data still centralizes shape validation and
```
harrisonritz, line 882: we're saving SLDS inputs for a future PR, eh?
Fair enough, already packed

harrisonritz (approved):
looks good!

src/lds/types.jl
```diff
@@ -20,14 +20,28 @@ abstract type AbstractObservationModel{T<:Real} end
...
+struct Data{
+    T<:Real,
+    YV<:AbstractVector{<:AbstractMatrix{T}},
+    UXV<:AbstractVector{<:AbstractMatrix{T}},
+    UYV<:AbstractVector{<:AbstractMatrix{T}},
+}
+    y::YV
+    ux::UXV
+    uy::UYV
+    tsteps::Vector{Int}
```
harrisonritz, line 46: Might as well add all the relevant shape info here.

Where is the constructor?

src/utils/validation.jl
```diff
@@ -541,9 +545,76 @@ end
...
+`nothing` when the model has no `B` / `D` input matrix; absent inputs are
+canonicalized to zero-row matrices.
+
+# Throws
+- `DimensionMismatchError` when observation or input dimensions disagree with
+  the model, or input trial lengths disagree with `y`
+- `ArgumentError` when inputs are omitted for a model that requires them
+  (`ux_dim > 0` / `uy_dim > 0`)
+"""
+function Data(
```
harrisonritz: Oh here is the Data constructor -- your call, but maybe put in types? eg could add a minimal internal constructor.

harrisonritz (approved):

src/lds/fit_SLDS.jl
rsenne: yes this will be next PR

src/utils/validation.jl
rsenne: gonna save this for #137

src/lds/fit_LDS.jl
rsenne: havent ran into this but the change is minimal so i see no downside in doing it for the sake of safety

src/lds/fit_SLDS.jl
harrisonritz: good call.

harrisonritz:
hooray! 🎉

PR #154, opened 2026-07-18
Feature/plds inputs

rsenne:
Adds PLDS input support on both the state and observation side


src/stats/simulate.jl
```diff
@@ -76,7 +84,16 @@ function _sample_trial!(
             ),
         )
         y_trial[:, t] =
-            rand.(rng, Poisson.(exp.(obs_params.C * x_trial[:, t] + obs_params.d)))
+            rand.(
```
harrisonritz, line 87: This is horrific, but better to be consistent with the linter

test/Project.toml
```diff
@@ -16,6 +16,7 @@ Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+SSDTest = "8a6b4d2f-3a7e-4d1c-9c5e-7a8b3f4d6c12"
```
harrisonritz: You might have already fixed by the time you see this, but think this was killing the check

harrisonritz (approved):

src/stats/simulate.jl
rsenne: Luckily i think JuliaFormatter 2.10 is stable...maybe we can migrate soon lol

PR #157, opened 2026-07-21
Make Initial State Shared in SLDS

rsenne:
@harrisonritz Can you tell me if this fixes your SLDS woes?

### DIFF
--- a/src/lds/fit_SLDS.jl
+++ b/src/lds/fit_SLDS.jl
@@ -692,6 +692,9 @@ function _slds_prior_logdensity(slds::SLDS{T}) where {T<:Real}
         if sm.P0_prior !== nothing
             prior_term += iw_logprior_term(sm.P0, sm.P0_prior)
         end
+        if sm.x0_prior !== nothing
+            prior_term += mn_logprior_term(reshape(sm.x0, :, 1), sm.P0, sm.x0_prior)
+        end
         if sm.AB_prior !== nothing
             ux_dim = lds.ux_dim
             W_ab = Matrix{T}(undef, D, D + 1 + ux_dim)
--- a/src/lds/fit_SLDS.jl
+++ b/src/lds/fit_SLDS.jl
@@ -892,6 +892,16 @@ function mstep!(
     # weighted aggregator.
     suf = _initialize_td_sufficient_statistics(T, slds.LDSs[1], tsteps_per_trial)
 
+    #=
+    x0/P0 are tied across modes. Since the smoother gives one q(x) per trial
+    and Σₖ γₖ(t=1) = 1, the pooled unit-weight init stats are exactly the sum
+    over modes of the per-mode init stats the aggregator already computes —
+    so accumulate them across the loop instead of recomputing.
+    =#
+    D = slds.LDSs[1].latent_dim
+    init_xy = zeros(T, 1, D)
+    init_yy = zeros(T, D, D)
+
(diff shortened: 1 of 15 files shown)


src/lds/workspaces.jl
```diff
@@ -97,9 +97,10 @@ mutable struct SufficientStatistics{T<:Real}
     without truncation.
     =#
     init_n::T
-    init_xx::Base.RefValue{DensePDMat{T}}
     init_xy::Matrix{T}
-    init_yy::Base.RefValue{DensePDMat{T}}
+    #= Raw scatter `Σγ(x₁x₁' + P₁)`, only ever read as a plain matrix. Not a PDMat:
+    a zero-weight regime makes it the zero matrix (would fail the Cholesky). =#
+    init_yy::Base.RefValue{Matrix{T}}
```
harrisonritz, line 103: Ya I think this is the right call. High risk for degeneracy issues, low computational demands, accuracy will matter less at longer tstep. 

FYI lots of discussion about how to treat initial conditions in the [Holmes tutorial](https://arxiv.org/abs/1302.3919) (eg section 3.8), though I don't think they are considering epoched data.

src/lds/fit_SLDS.jl
```diff
@@ -901,26 +911,52 @@ function mstep!(
...
+    slds::SLDS{T}, suf::SufficientStatistics{T}, sws::SmoothWorkspace{T}
+) where {T<:Real}
+    lds1 = deepcopy(slds.LDSs[1])
+    update_initial_state_mean!(lds1, suf)
+    update_initial_state_covariance!(lds1, suf, sws)
+    fit_x0, fit_P0 = lds1.fit_bool[1], lds1.fit_bool[2]
+    for k in eachindex(slds.LDSs)
+        fit_x0 && copyto!(slds.LDSs[k].state_model.x0, lds1.state_model.x0)
+        fit_P0 && copyto!(slds.LDSs[k].state_model.P0, lds1.state_model.P0)
+    end
```
harrisonritz, line 959: Feel free to tweak my change here -- k=1 would crash, I think, and k>1 would accidentally update lds[1] even when fit_bool=false

src/lds/fit_SLDS.jl
```diff
@@ -901,26 +911,52 @@ function mstep!(
...
+from the pooled init stats in `suf` (see `mstep!`) and copy it into every
+`state_model`.
+"""
+function _update_shared_initial_state!(
+    slds::SLDS{T}, suf::SufficientStatistics{T}, sws::SmoothWorkspace{T}
+) where {T<:Real}
+    lds1 = deepcopy(slds.LDSs[1])
+    update_initial_state_mean!(lds1, suf)
+    update_initial_state_covariance!(lds1, suf, sws)
+    fit_x0, fit_P0 = lds1.fit_bool[1], lds1.fit_bool[2]
```
harrisonritz, line 955: Hard indices on fit_bool make me nervous, at some point we should make it a named tuple or something.

harrisonritz (approved):

src/lds/fit_SLDS.jl
rsenne: oh yeah i despise `fit_bool` it was the classic "nothing more permanent than a temporary solution`

src/lds/fit_SLDS.jl
rsenne: this is perfect thanks!

PR #161, opened 2026-08-11
Add `posterior` function to estimate MLE/MAP elbo, responsibility, and smoothed states

harrisonritz:
Iterates discrete/continuous smoothing until convergence, then exports elbo, responsibilities, and states. Takes a while to run to completion, so ideally for post-fit eval.

- want a better interface to choose what is exported -- maybe an `export` dict?

- export smoothed cov, but make it optional (big variable)

- **remove kluge regularization of discrete parameters**

- maybe we create a function to share with the estep? Eg could have option to run multiple smoothing iterations during estep (iirc you said 2-3 iter can give much (?) better updates). Less surface for estep revisions


rsenne:
I'll do my best to review this today, thought it's a big diff so may take me a day or two :) thanks!

rsenne:
i will review this by EOD

rsenne:
> kluge regularization 

wtf is this lol

test/runtests.jl
```diff
@@ -111,6 +111,11 @@ using SSDTest
                 test_SLDS_estep_elbo_components()
                 test_SLDS_elbo_matches_LDS_marginal_K1()
                 test_SLDS_public_elbo()
+                @skip_test test_SLDS_posterior_basic()
+                @skip_test test_SLDS_posterior_shapes()
+                @skip_test test_SLDS_posterior_deterministic_and_modes()
+                @skip_test test_SLDS_posterior_K1()
+                @skip_test test_SLDS_posterior_recovers_distinct_regimes()
```
rsenne: probably wanna unskip these?

test/LinearDynamicalSystems/SLDS.jl
```diff
@@ -1488,6 +1488,161 @@ function test_SLDS_fit_shapes_and_validation(; rng=MersenneTwister(0xE1B1))
...
+    obs_dim = 3
+    tsteps = 15
+    ntrials = 2
+
+    lds = _make_gaussian_lds_dense(latent_dim, obs_dim; seed=3)
+    slds = SLDS(; A=_rowstochastic(K), πₖ=_probvec(K), LDSs=[lds, deepcopy(lds)])
+    z, x, y = rand(rng, slds, fill(tsteps, ntrials))
+
+    # Single-trial matrix ⇒ bare K × T matrix; wrapping in a vector ⇒ 1-element vector.
+    γ_mat = infer_γ(slds, y[1])
```
rsenne: many instances of this stale function handle

docs/src/SLDS.md
```diff
@@ -91,6 +91,25 @@ The sampling process follows the generative model:
...
+responsibilities converge or for a fixed number of iterations. This interleaving of
+discrete and continuous smoothing within the E-step follows the classic
+coordinate-ascent scheme of Ghahramani & Hinton (1996).
+
+Unlike the single-Monte-Carlo-sample E-step that `fit!` uses during learning, the
+coupling here is deterministic (the discrete-layer log-likelihoods are evaluated at
+the smoothed posterior mean), so the returned responsibilities are reproducible.
+
+```@docs
+infer_γ
```
rsenne: needs to be updated

src/lds/fit_SLDS.jl
```diff
@@ -99,24 +141,36 @@ function Random.rand(
...
 end
 
-function Random.rand(slds::SLDS, tsteps_per_trial::AbstractVector{<:Integer})
-    return rand(Random.default_rng(), slds, tsteps_per_trial)
+function Random.rand(slds::SLDS, tsteps_per_trial::AbstractVector{<:Integer}; kwargs...)
+    return rand(Random.default_rng(), slds, tsteps_per_trial; kwargs...)
 end
 
-# Core SLDS trial sampling logic
+# Core SLDS trial sampling logic. `ux_trial` / `uy_trial` are the canonicalized
```
rsenne: seems liek this last predicate was meant to be deleted?

src/lds/fit_SLDS.jl
```diff
@@ -300,18 +395,22 @@ function joint_loglikelihood!(
...
 """
-    gradient!(ws, slds, x, y, w)
+    gradient!(ws, slds, x, y, w[, ux, uy])
 
 In-place SLDS gradient: each component's complete-data gradient is scaled
 per-timestep by the responsibility `w[k, t]` and accumulated. Writes into
-`ws.opt.grad_buf` and returns it.
+`ws.opt.grad_buf` and returns it. `ux` (dynamics input, feeds `-Q⁻¹` /
+`A'Q⁻¹` residuals via `Bₖ u`) and `uy` (observation input, feeds the emission
+gradient via `Dₖ v`) are per-trial matrices; `nothing` or zero-row skips them.
```
rsenne: I'm assuming claude added these types of statements. Probably can mention how inputs are done once and don't need to mention again unless the function is specific to them or has some behavior meant to truly notice

src/lds/fit_SLDS.jl
```diff
@@ -1174,3 +1378,272 @@ function fit!(
...
+- `progress::Bool=false`: show a progress bar.
+
+# Returns
+The responsibilities `γ`:
+- for a single-trial matrix `y`, a `K × T` matrix; and
+- otherwise (3-D array or vector-of-matrices input), a `Vector` of per-trial
+  `K × T_i` matrices, one per trial.
+
+Each column is a probability vector over the `K` discrete states (sums to 1).
+"""
```
rsenne: can we de-claude this?

src/lds/fit_SLDS.jl
```diff
@@ -1174,3 +1378,272 @@ function fit!(
...
+Core deterministic variational-E-step iteration behind [`posterior`](@ref). Operates
+on the canonicalized per-trial observation sequence `y_seq` (a `Vector` of
+`(obs_dim, T_i)` matrices) and returns one `K × T_i` responsibility matrix per
+trial. Holds the `SLDS` parameters fixed (no M-step); the per-regime constants are
+cached once and reused every iteration.
+"""
+function posterior(
+    slds::SLDS{T,S,O},
+    data::Data{T};
+    return_γ::Bool=true,
```
rsenne: if someone is calling posterior--is there really a circumstance in which they dont want gamma? I mean to an extent what sets this apart from smooth is really just the gammas

src/lds/fit_SLDS.jl
```diff
@@ -1174,3 +1378,272 @@ function fit!(
...
+  in any responsibility between successive iterations (used only when
+  `check_convergence=true`).
+- `check_convergence::Bool=true`: the mode flag. `true` iterates *until
+  convergence* (`max|Δγ| < tol`), capped at `max_iter` iterations; `false` runs
+  *exactly* `max_iter` iterations.
+- `progress::Bool=false`: show a progress bar.
+
+# Returns
+The responsibilities `γ`:
+- for a single-trial matrix `y`, a `K × T` matrix; and
```
rsenne: i don't think this is accurate

src/lds/fit_SLDS.jl
```diff
@@ -1174,3 +1378,272 @@ function fit!(
...
+    slds::SLDS{T,S,O},
+    y::AbstractMatrix{T};
+    ux=nothing,
+    uy=nothing,
+    return_γ::Bool=true,
+    return_elbo::Bool=true,
+    return_x::Bool=false,
+    max_iter::Int=100,
+    tol::Real=1e-6,
+    check_convergence::Bool=true,
```
rsenne: this also seems a bit unneccesary. I mean we probably should just check convergence and if we don't ust throw a warnign and return given the max_iters

src/lds/fit_SLDS.jl
```diff
@@ -1174,3 +1378,272 @@ function fit!(
...
+Each column is a probability vector over the `K` discrete states (sums to 1).
+"""
+function posterior(
+    slds::SLDS{T,S,O},
+    y::AbstractMatrix{T};
+    ux=nothing,
+    uy=nothing,
+    return_γ::Bool=true,
+    return_elbo::Bool=true,
+    return_x::Bool=false,
```
rsenne: I think there is a bit of semantic dubiousness going on. we noe effecticely have two ways to get elbo right? so why should this function privilege the "good elbo" ? I wonder if this function need to be integrated more into the core code somewhat like in ssm -- e.g., the same function does the posterior, but depending on the calls, we wi8ther iterate until convergence or do it once etc.

rsenne:
I think overall the functions are well written mopdulo some correctness issues. I do think the semantics need to be re-evaluated `posterior` is really just smoothing, so I think it might be more appropriate to get this machinery into `smooth` or a variant of it, and then make the public return both in a NamedTuple and then we can have the necessary kwargs to specify maxiters etc. thoughts?

harrisonritz:
Ok, so maybe `smooth` just needs to have an `smoothing_iters` argument (as well as `fit!`/`estep!`, for that matter). 

And then `smooth` can just return gamma, and the loglikelihood can return the elbo.

The only hiccup is that you might only want to run the back-and-forth estimator once -- because it might take many iterations to converge -- so nice to have it spit out the posterior and the elbo from the same fit. Maybe some caching would help here.

rsenne:
> smooth just needs to have an smoothing_iters

yeah that should work. In theory the `smooth!` variant can just default to 1 iteration to stay fiathul to OG vLEM algo

> And then smooth can just return gamma

do you mean gamma and x? also i don't mind too much if return the elbo as part of a named tuple in `smooth`. my biggest concern is two places to compute the elbo and the one not called `elbo` is actually the more accurate one. we should probably also have some way in elbo to allow for the approximation to be better via more iters

harrisonritz:
Ok, I'll take a look next week!

Meant smooth should now additionally return gamma (and elbo, good idea).

harrisonritz:
ok, I took another pass on this. 
You can see a description of the revision in https://github.com/depasquale-lab/StateSpaceDynamics.jl/pull/175 

one important decision is whether to resample q(x) each iteration, or whether to keep using the mean from the previous iteration. Here is the description from the new ` _vem_alternate!` function (modularized and used across `fit!` and `smooth`):

```
Run `smoothing_iters` discrete↔continuous alternations of the structured-variational
E-step. One alternation is:

1. fill `dl.logL` (`K × sum(T_i)`) with per-regime log-likelihoods of the current
   continuous trajectory,
2. refresh q(z) = γ by forward-backward over the switching chain (HMMs.jl threads
   across trials), and
3. refresh q(x) by re-running the Laplace/Newton smoother on each trial under the new
   γ, filling `tfs[*].x_smooth`, `tfs[*].p_smooth`, and `tfs[*].entropy`.

`x_samples` selects how step 1 reads the continuous trajectory, and is the only
difference between the two callers:

- `x_samples === nothing` — plug in the smoothed mean `E_q[x]`. Deterministic and
  reproducible; used by [`smooth`](@ref) for post-fit inference.
- `x_samples !== nothing` — plug in a joint draw from q(x), and draw the next one in
  step 3. This is the vLEM Monte-Carlo E-step used by [`fit!`](@ref); `x_samples` is
  read then overwritten within each alternation.
  ```
  
It keeps the biased version (use previous draw) in `fit!` so that it reproduces the original behaviour. 

However, it defaults to using the unbiased version in `smooth` (`x_samples === nothing`), which might be slower to converge. Take a look at the implementation, something worth thinking about -- you have better intuitions about stochastic VEM than I do.

**Comment by rsenne** [comment]
I''ll have review / updates on thsi today

**Comment by rsenne** [comment]
seems we have a flaky _broken_ test?

**Comment by rsenne** [comment]
@harrisonritz want to review the changes i made?

File: src/lds/types.jl [hunk 11]
```diff
@@ -17,6 +17,24 @@ data are generated from the latent state.
...
+- `AbstractVector{<:AbstractMatrix{T}}` - one `(obs_dim, T_i)` matrix per trial,
+  ragged lengths allowed.
+
+**Internal** alias, used only to keep the signatures of `fit!`, `smooth`,
+`elbo` and `loglikelihood` readable. `Data(model, y; ux, uy)` normalizes any of
+the three into the per-trial form everything downstream assumes.
+"""
+const Observations{T} = Union{
+    AbstractMatrix{T},AbstractArray{T,3},AbstractVector{<:AbstractMatrix{T}}
+}
```
harrisonritz, line 36: this is a great idea! 
Its a little extraneous to the PR, but its small enough that lets just keep it

src/lds/fit_SLDS.jl
```diff
@@ -1319,24 +1308,8 @@ end
...
-  side updates flow through the same suf path; the emission [C d] is updated
-  via the existing LBFGS routine (Poisson is non-conjugate and cannot be
-  folded into the regression).
-
-When inputs are present the weighted aggregator folds `Bₖ u` / `Dₖ v` into the
-regression targets, so `Bₖ` and `Dₖ` are re-estimated alongside `Aₖ` / `Cₖ`.
+Update the transition probabilities and each regime's parameters from
+γ-weighted sufficient statistics. Poisson emissions use their LBFGS update.
 """
 function mstep!(
```
harrisonritz: is all this info documented elsewhere? Isn't the website generated from this docstring?

harrisonritz (commented):
Observations type is a great idea, let's keep it.
You trimmed a lot of the mstep docstring, just want to make sure thats documented elsewhere.

harrisonritz:
> seems we have a flaky _broken_ test?

wait so `lts` is failing because it passes? thats weird

src/lds/fit_SLDS.jl
harrisonritz: ehh nvm you're probably right, this isn't all needed

rsenne:
> However, it defaults to using the unbiased version in smooth

i'm a bit confused by this. using the mean as a plug in estimator would be biased whereas a sample would not be?

rsenne:
this is my last remaining concern related to this PR^

rsenne:
@harrisonritz any further thoughts?

harrisonritz:
looks good!

rsenne (approved):

Issue #98, opened 2026-07-01
Decide how parameters are handled internally.


rsenne:
E.g., how we handle the cases where we want a bias/not a bias vs input etc. See #94 for lots of discussion on this point.

harrisonritz:
There is a related issue of how we  link properties to parameters that I've been thinking a lot about. 

Eg

- whether a parameter depends on discrete state
- which discrete state a parameter corresponds to
- whether a parameter is fit at all
- whether a parameter differs across trials (eg session-specific C matrices for stitching)
- which parameters correspond to which trials
- ???

One method would be to build out dyn/obs classes, but that might grow exponentially. Sidecar-ing (eg `fit_bool`) can do some, but it's easy to make mistakes.

**Proposed solution**: parameter-level types.

Here's a trial-varying examples I've been clauding. We'd make types for trial-static and trial-varying parameters:

```julia

abstract type Param{T} end

@kwdef struct Static{T,A} <: Param{T}
     value::A
     frozen::Bool = False
end  

@kwdef struct TrialVary{T,A} <: Param{T}
     values::Vector{A}
     trial_index::Vector{Int}
     frozen::Bool = False
end

Static(v; frozen=false) = 
     Static{eltype(v), typeof(v)}(value=v, frozen=frozen)

TrialVary(vs::Vector, trial_index; frozen=false) =
     TrialVary{eltype(eltype(vs)), eltype(vs)}(values=vs, trial_index =index, frozen=frozen)
```

Can set parameters:

```julia

lds.A = Static(values=randn(10,10))
lds.C = TrialVary(
         values=[randn(20,10) for _ in 1:2], 
         trial_index=repeat([1, 2], inner=50)
         )

```

And then can call parameters with something like this in functions like `smooth!`, which already extracts the parameters.

```julia
at(p::Static, k) = p.value
at(p::TrialVary, k) = p.values[p.trial_index[k]]

C = at(lds.C, 51)

```

Maybe could have `TrialVary`, `TimeVary`, and `TimeTrialVary` types, sharing an `at()` interface.

**Summary**: would be nice for code to handle flexible container types (eg matrix, vector-of-matrices) for time-/trial-varying parameters, and instead of building more dynamics/observation classes, maybe could put more into into the parameter itself (ie by making it a struct). Eager to hear your thoughts.

rsenne:
[...]

PR #165, opened 2026-08-13
feature / condition-varying parameters

harrisonritz:
Every AbstractStateModel and AbstractObservationModel gains a `depends_on` field (default `nothing`). Setting it to a NamedTuple of per-trial label vectors makes the named parameters be estimated separately for each group of trials, while everything else stays pooled:

    obs.depends_on = (C = session, R = session)

That is the stitching setup for combining recording sessions that observe different neurons in the same animal: latent dynamics shared by the animal, emission parameters specific to the session.

Design. A *cell* is a maximal set of trials sharing every parameter — the common refinement of all the label vectors. Trials in a cell are governed by a single LinearDynamicalSystem, so the E-step runs on each cell's sub-`Data` completely unchanged; only the M-step and the ELBO become group-aware. This keeps the equal-length shared-covariance fast path alive *within* each group (parameters differ between cells, so their covariances genuinely differ), and the O(D^2*T) workspace storage is allocated once and reused across cells, so a grouped fit's memory tracks an ungrouped one's rather than scaling with the number of sessions.

Keys canonicalize to the same groups fit_bool uses, since those parameters are fit jointly as one regression: :A/:b/:B name one group, :C/:d/:D another. Labels may be Symbols, integers or strings. Different parameters may use different label vectors.

Per-group values live in a new `variants` field holding one model object per parameter-group combination, with non-varying parameters shared by reference so a single M-step write covers all of them. They are read back with the new exported `group_labels` and `group_parameter`.

The covariance updates accumulate their residual scatter cell by cell, each with its own regression matrix, so a grouping of (say) C finer than that of R comes out right; the MN/IW prior terms are counted once per distinct parameter version. `update_Q!`, `update_R!` and `update_initial_state_covariance!` were refactored into accumulate/finalize halves so the grouped and ungrouped paths share one implementation.

Supported for the Gaussian LDS, the Poisson LDS and the SLDS across `fit!`, `smooth`, `elbo`, `loglikelihood` and `rand`, each of which also accepts a `depends_on` keyword overriding the model's stored labels so a held-out set with a different trial count can be scored without mutating the model. For an SLDS every regime must declare the same labels (the trial partition is a property of the data); x0/P0 stay tied across regimes.

With `depends_on` unset, every entry point takes its original code path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,41 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Ancillary parameter dependencies: every `AbstractStateModel` and
+  `AbstractObservationModel` now carries a `depends_on` field (default
+  `nothing`). Setting it to a `NamedTuple` of per-trial label vectors — e.g.
+  `obs_model.depends_on = (C = session, R = session)` — makes those parameters
+  be estimated separately for each group of trials, while everything else stays
+  pooled. This is the "stitching" setup for combining recording sessions that
+  observe different neurons in the same animal: shared latent dynamics,
+  session-specific `C`, `d`/`D` and `R`. Supported for the Gaussian LDS, the
+  Poisson LDS and the SLDS, across `fit!`, `smooth`, `elbo`, `loglikelihood`
+  and `rand`.
+  * Keys are canonicalized to the same groups `fit_bool` uses, since those
+    parameters are fit jointly as one regression — `:A`/`:b`/`:B` name one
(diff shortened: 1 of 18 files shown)


harrisonritz:
Discussed in https://github.com/depasquale-lab/StateSpaceDynamics.jl/issues/98

solution here is to combine parameters into cells, and then run each cell through the standard fitting procedure

rsenne:
Can we break this up into separate PRs? Its a bit big as is

harrisonritz:
> Can we break this up into separate PRs? Its a bit big as is

You probably saw, but now split up into multiple PRs: #166, #167, #168, #169, #170, #172

They are passing (besides a JET issue that should be addressed #173), ready for review at your leisure

